PR #1, opened 2016-03-11
Selenium driver inside jibri

aaronkvanmeerten:
also improved output with logging.info instead of print


aaronkvanmeerten:
Updated to include latest fixes for jibri

PR #6, opened 2016-06-01
Ensure that the PID file is always closed

SamWhited:
Fixes a minor resource leak where the PID file could end up not being closed. If this were to happen, the program probably would have crashed anyways, so it wouldn't matter, but since we don't know what kind of error handling is happening outside of the `writePidFile` function (that might be catching the error and not terminating the program) we should ensure that the file handle gets closed anyways.


aaronkvanmeerten:
Looks great!

PR #8, opened 2016-06-03
Python Packaging

SamWhited:
Refactored Jibri a bit to allow building a Python package. This probably needs more testing, as I have no idea if it will break any existing imports in prod or not.

This would allow us to push Jibri to PyPi, and it will create executables for the various commands when you install it.


saghul:
Things have changed too much so this no longer applies... in addition the setup is now documented and installing the Python package is one of the smaller tasks, so I'm closing this.

PR #9, opened 2016-06-07
Add Dockerfile and gitignore

SamWhited:
I was just talking to @emcho about possibly creating a Dockerfile for Jibri so I thought I'd go ahead and put this in PR for review.

I'm not entirely sure how things will work since you'll need the ALSA loopback kernel modules on your host machine. There's a chance we could run a different kernel in userspace in the container, but this would also probably limit the image to be Linux-only (and might make things really slow). Maybe someone who knows more about Docker and LXC than I do could chime in.

I'll start testing this tomorrow and see if I can't get it working.


SamWhited:
Current issues:

* Client fails to find scripts (maybe we need a `SCRIPTS_DIR` environment variable?)
* Hangs on "Initializing Driver"

cmoreira-daitan:
@SamWhited by the time you were developing/testing, icewm was available for Alpine? I can't find it anywhere anymore

SamWhited:
@cmoreira-daitan I don't recall what I was messing with; I think I swapped it out for something else. Probably i3.

PR #17, opened 2016-08-15
trim all incoming parameters to jibri process

aaronkvanmeerten:
Make sure input has been sanitized a little better


jibri-xmpp-client/app.py
```diff
@@ -260,6 +260,11 @@ def jibri_start_callback(client, url, stream_id, room=None, token='token', backu
     global selenium_xmpp_password
 
 
+    #make sure we remove whitespace from all input parameters
+    room.strip()
```
bgrozev: I don't think this works, I think you need something like "room = room.strip()"

aaronkvanmeerten:
Thanks for the comment, I've fixed it

PR #23, opened 2016-12-21
Fix potential bug in waitXMPPConnected()

debugmaster:
waitXMPPConnected() checks the XMPP status by calling itself and its timeout check is never reached. 
Change waitXMPPConnected() to isXMPPConnected() to fix it.


aaronkvanmeerten:
Awesome catch thanks :)

PR #39, opened 2017-06-07
Running multiple Jibri instances

krakatoa:
Hi guys, I've been doing some tests with multiple Jibri instances running on a single host, and it seems pretty stable. I wanted to share it here, any feedback will be great.
Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,28 @@ Once recording is enabled in config.js, the recording button will become availab
    - check your youtube livestreaming dashboard for the status (https://youtube.com/live_dashboard)
    - `./stop_recording.sh` to kill the stream
 
+# Running more than one Jibri per host
+
+* To have multiple instances of ALSA Loopback, add the following lines to the `/etc/modprobe.d/alsa-base.conf` (this adds 5 loopback devices):
+
+  ```
+  alias snd-card-0 snd-aloop
+  options snd-aloop enable=1,1,1,1,1 index=0,1,2,3,4
+  ```
+
+  For each card, add a mixer into the `.asoundrc`. This is a sample block for a card with index 3:
+
+  ```
+  pcm.amix_3 {
+    type dmix
+    ipc_key 219345
+    slave.pcm "hw:3,0,0"
+  }
+  ```
+
+* Run as many Xorg instances as required, each with a different DISPLAY id.
+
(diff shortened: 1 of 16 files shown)


saghul:
Hi there! This looks really nice! I took a quick peek and nothing stood out as weird. You also get the extra credit for adding some docs!

@aaronkvanmeerten WDYT?

krakatoa:
@saghul thanks for reviewing it! I found a glitch however, the scripts/stop_selenium.sh is using `killall -9` for both chrome & chromedriver, I'll need to find a way to deal with them without turning other Selenium instances off. Will check it and be back to you.

aaronkvanmeerten:
Great work here.  I think that your best bet here will be to somehow identify which chromedriver was instantiated with each jibri session, and then use that PID to determine what are the child PIDs of chrome itself.  Then instead of using killall, we could kill each of the specific PIDs individually.  I'd be happy to discuss this with you more if you're interested.

krakatoa:
@aaronkvanmeerten @saghul glad you guys liked the changes :+1: I already fixed the `killall` part, and after testing it several times on my side, I'm confident that a Selenium crash on Jibri A doesn't take down Jibri B. Can you review it please?
Thank you!

aaronkvanmeerten:
Out of curiosity, have you tried recording many calls simultaneously from a single jibri instance?  One of the reasons we didn't push too hard for that is when receiving/rendering many streams, CPU utilization on a reasonably sized instance can go to 80% even for a single HD call.  So in order to preserve quality we plan the recording instances to be sized for individual calls.

krakatoa:
Glad to collaborate in the community !
And no, I haven't tried recording many calls in a single jibri instance, as I understand that the code will require lots changes (there are unique global locks, single per-action callback watchers, handling of individual PIDs to track external processes). It becomes easier to spin a separate process for each Jibri. If you ask me, it would always be nice to have Jibri prepared to deal with multiple concurrent task listeners (leaving this option up to the user), so let's suppose I find that using a GPU powered host, or a huge CPU one gives me the room to run several Jibris, that would be ideal.

bbaldino:
@krakatoa the new java-based jibri has been merged so you'll see quite a few conflicts here.  have you looked at the new version at all and considered moving to it? it would be great to have this work updated to the new version, as the old (python) one won't be maintained anymore.

lggwettmann:
Is there any chance this pull-request can be moved forward somehow? Would be fantastic if we could enable multiple jibri instances on one server!

bbaldino:
I'm going to close this, as it's not long outdated (since it's based on the old jibri), but there is work on a docker-based jibri container [here](https://github.com/jitsi/docker-jitsi-meet)

PR #45, opened 2017-09-12
File recording

aaronkvanmeerten:
This is a pretty big PR based on experimental work on file recording.  The idea is to enable a custom deployment to override the scripts/finalize_recording.sh to perform the custom copy and notification actions desired.

This also introduces mp4 wrapping for file recordings, as well several tunings to the ffmpeg process.  The biggest of these tunings is also the move to AAC audio for all files AND livestreams.

ffmpeg is now initially stopped with a kill -2 instead of a kill -9, to allow more graceful shutdown processing and final writing to any files of streams.

### DIFF
--- a/jibri-xmpp-client/app.py
+++ b/jibri-xmpp-client/app.py
@@ -619,7 +619,7 @@ def launch_pjsua(sipaddress, displayname=''):
         return 'pjsua_startup_streaming_exception'
 
 def launch_ffmpeg(url,recording_path='',token='',stream_id='', backup=''):
-    recording_file = recording_path + datetime.now().strftime('/%Y%m%d%H%M%S.flv')
+    recording_file = recording_path + datetime.now().strftime('/%Y%m%d%H%M%S.mp4')
     #we will allow the following number of attempts:
     try:
         #first try to start ffmpeg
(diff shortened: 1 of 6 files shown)


jibri-xmpp-client/app.py
```diff
@@ -315,6 +329,18 @@ def jibri_start_callback(client, url, stream_id, sipaddress=None, displayname=No
     stream_id=stream_id.strip()
     sipaddress=sipaddress.strip()
 
+    #assume we're streaming if a stream_id is provided
+    if stream_id:
+        recording_mode='stream'
+
+    #default to file-based recording if none is specified
+    if not recording_mode:
```
bbaldino: haven't you already covered this with the default param?

jibri-xmpp-client/app.py
```diff
@@ -315,6 +329,18 @@ def jibri_start_callback(client, url, stream_id, sipaddress=None, displayname=No
...
+        recording_mode = 'file'
+
+    #clear stream id if file mode is set
+    if recording_mode == 'file':
```
bbaldino: did you consider treating this as an error?  (setting `stream_id` when `recording_mode` is `file`) wonder if it might be clearer

PR #58, opened 2017-11-21
Updates around sipgw improvements

damencho:


jibri-xmpp-client/jibrixmppclient.py
```diff
@@ -322,11 +334,13 @@ def report_jibri_error(self, error):
         except Exception as e:
             logging.error("Failed to send status update: %s", str(e))
 
-    def update_jibri_status(self, status):
+    def update_jibri_status(self, status, sipaddress):
```
aaronkvanmeerten: could provide a default value for sipaddress here so that it doesn't require the parameter in all calls

PR #72, opened 2018-03-13
new jibri

bbaldino:


resources/finalize_recording.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+RECORDINGS_DIR=$1
+
+echo "This is a dummy finalize script" > /tmp/finalize.out
+echo "The script was invoked with recordings directory $RECORDINGS_DIR." >> /tmp/finalize.out
+echo "You should put any finalize logic (renaming, uploading to a service" >> /tmp/finalize.out
+echo "or storage provider, etc.) in this script" >> /tmp/finalize.out
+
+exit 0
```
virtuacoplenny: Why does this dummy script exist?

src/main/java/org/jitsi/jibri/CallUrlInfo.kt
```diff
@@ -0,0 +1,27 @@
...
+data class CallUrlInfo(
+        val baseUrl: String = "",
+        val callName: String = ""
+) {
+    val callUrl = "$baseUrl/$callName"
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null) return false
+        if (javaClass != other.javaClass) return false
```
virtuacoplenny: Where does "javaClass" come from?

src/main/java/org/jitsi/jibri/JibriManager.kt
```diff
@@ -0,0 +1,203 @@
...
+        ))
+        return startService(service, serviceParams, serviceStatusHandler)
+    }
+
+    /**
+     * Helper method to handle the boilerplate of starting a [JibriService].
+     * Returns a [StartServiceResult] to denote whether the service was
+     * started successfully or not.
+     */
+    private fun startService(
```
virtuacoplenny: Pardon the ignorance, what stops multiple places from calling startService at the same time? If that's possible, is that a problem?

src/main/java/org/jitsi/jibri/JibriManager.kt
```diff
@@ -0,0 +1,203 @@
...
+        // The manager adds its own status handler so that it can stop
+        // the error'd service and update presence appropriately
+        jibriService.addStatusHandler {
+            when (it) {
+                JibriServiceStatus.ERROR, JibriServiceStatus.FINISHED -> stopService()
+            }
+        }
+
+        if (!jibriService.start()) {
+            return StartServiceResult.ERROR
```
virtuacoplenny: Does something at this point need to publish that the status is no longer BUSY?

src/main/java/org/jitsi/jibri/Main.kt
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+        val jerseyConfig = ResourceConfig()
+        jerseyConfig.register(internalHttpApi)
+            .register(ContextResolver<ObjectMapper> { ObjectMapper().registerModule(KotlinModule()) })
+            .register(JacksonFeature::class.java)
+
+        val servlet = ServletHolder(ServletContainer(jerseyConfig))
+
+        val server = Server(3333)
```
virtuacoplenny: Is it known to be safe to have these ports be hardcoded?

src/main/java/org/jitsi/jibri/api/http/HttpApi.kt
```diff
@@ -0,0 +1,93 @@
...
+                    ServiceParams(usageTimeoutMinutes = 0),
+                    StreamingParams(serviceParams.callParams, serviceParams.youTubeStreamKey
+                    )
+                )
+            }
+        }
+        return when (result) {
+            StartServiceResult.SUCCESS -> Response.ok().build()
+            StartServiceResult.BUSY -> Response.status(Response.Status.PRECONDITION_FAILED).build()
+            StartServiceResult.ERROR -> Response.status(Response.Status.INTERNAL_SERVER_ERROR).build()
```
virtuacoplenny: It's okay there is no default cases?

src/main/java/org/jitsi/jibri/api/http/HttpApi.kt
```diff
@@ -0,0 +1,93 @@
+package org.jitsi.jibri.api.http
```
virtuacoplenny: Someone once yelled at me for not having licenses in my files.

src/main/java/org/jitsi/jibri/CallUrlInfo.kt
paweldomas: what is .kt extension ? kotlin ?

virtuacoplenny: Yes, jibri got re-written in kotlin.

paweldomas:
One of the arguments for rewriting Jibri in Java was that we already have ways for working around the webdriver issues in jitsi-meet-torture. Are we sharing any code with torture or are there any plans to do so ? Does it even make any sense to do that ?

bbaldino:
> One of the arguments for rewriting Jibri in Java was that we already have ways for working around the webdriver issues in jitsi-meet-torture. Are we sharing any code with torture or are there any plans to do so ? Does it even make any sense to do that ?

Which code are you referring to here?  I'd be all for that, I think I just wasn't aware of it.  If you're wondering about reuse because this is kotlin, kotlin is fully compatible with java so we can still reuse code there (and in fact this already does reuse some things, like the `JibriIq` class and some other xmpp-related code)

src/main/java/org/jitsi/jibri/Main.kt
bbaldino: no, this is hacked for now.  i have a todo list item to make the ports configurable

paweldomas:
>Which code are you referring to here? I'd be all for that, I think I just wasn't aware of it. If you're wondering about reuse because this is kotlin, kotlin is fully compatible with java so we can still reuse code there (and in fact this already does reuse some things, like the JibriIq class and some other xmpp-related code)

Yes, I know that kotlin will work for that. I can not refer to any particular code here, as I haven't read it all yet, but my guess would be things around dealing with the selenium driver, anything around interaction with the web app through the driver (if there is any).

bbaldino:
> Yes, I know that kotlin will work for that. I can not refer to any particular code here, as I haven't read it all yet, but my guess would be things around dealing with the selenium driver, anything around interaction with the web app through the driver (if there is any).

Yeah I agree...I didn't think we had any pageobjects in java for the web so i added some basic ones here...it would be great to get those in a central place so they could be used by everything that needs them.

bgrozev:
I haven't read this carefully (I will try to take a closer look later), but a couple of observations:
Since we're now using Kotlin, I think we should agree on some code conventions and follow them everywhere. I would suggest to adopt the java convention wherever appropriate, as kotlin and java code are going to appear together often. What do we do about:
* licence headers
* line length
* imports (by-class vs *)
* braces

My other questions is about the xmpp/muc code. Are we still planning to have a common base in jicoco that we can reuse in both jitsi-videobridge and jibri?

src/main/java/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,214 @@
...
+                    .setXmppDomain(config.controlLogin.domain)
+                    .setUsernameAndPassword(config.controlLogin.username, config.controlLogin.password)
+                    .build())
+                mucClient.addIqRequestHandler(object : JibriSyncIqRequestHandler() {
+                    override fun handleJibriIqRequest(jibriIq: JibriIq): IQ {
+                        return handleJibriIq(jibriIq, config, mucClient)
+                    }
+                })
+                jibriManager.addStatusHandler { status ->
+                    logger.info("XMPP API got jibri status $status, publishing presence to connection ${config.name}")
```
virtuacoplenny: Is there any config validation? Is that even needed?

src/main/java/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,214 @@
...
+            resultIq.to = startJibriIq.from
+            resultIq.type = IQ.Type.set
+            try {
+                logger.info("Starting service")
+
+                // If there is an issue with the service while it's running, we need to send an IQ
+                // to notify the caller who invoked the service of its status, so we'll listen
+                // for the service's status while it's running and this method will be invoked
+                // if it changes
+                val serviceStatusHandler: JibriServiceStatusHandler = { serviceStatus ->
```
virtuacoplenny: I think this was the function with the 9 cyclomatic complexity. Just based on the level of nesting, and the normal level of nesting you've been aiming for, I think this one stands out as a "beast" compared to the lightweight helpers you've been creating. I could see the function being broken out into a separate helper (and thereby potentially becoming easier to test) and maybe...enum/mapping for the stats?

src/main/java/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,214 @@
...
+                    when (serviceStatus) {
+                        JibriServiceStatus.ERROR -> {
+                            logger.info("Current service had an error")
+                            val errorIq = JibriIq()
+                            errorIq.to = startJibriIq.from
+                            errorIq.type = IQ.Type.set
+                            errorIq.status = JibriIq.Status.FAILED
```
virtuacoplenny: I'm seeing this pattern a lot where JibriIq() creates something and then you manually set values on it. Is this stylistically how attributes are normally set? I'd be more used to passing these into the factory function to potentially hide some kind of implementation, if it even needs hiding.

src/main/java/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,214 @@
...
+            } finally {
+                logger.info("Sending start service response iq: $resultIq")
+                mucClient.sendStanza(resultIq)
+            }
+        }
+        // Immediately respond that the request is pending
+        val initialResponse = JibriIqHelper.createResult(startJibriIq)
+        initialResponse.status = JibriIq.Status.PENDING
+        logger.info("Sending 'pending' response to start IQ")
+        return initialResponse
```
virtuacoplenny: By the time I got here I had forgotten what function I was in. I think that's a sign to pull out the middle stuff.

src/main/java/org/jitsi/jibri/capture/ffmpeg/FfmpegCapturer.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val logger = Logger.getLogger(this::class.qualifiedName)
+    val ffmpegExecutor: FfmpegExecutor
+
+    init
+    {
+        logger.debug("OS: ${System.getProperty("os.name")}")
+        ffmpegExecutor = when (System.getProperty("os.name")) {
+            "Mac OS X" -> MacFfmpegExecutor()
+            "Linux" -> LinuxFfmpegExecutor()
+            else -> throw UnsupportedOsException()
```
virtuacoplenny: I'm wondering if it's best to hide this completely in another factory so FfmpegCapturer doesn't have to know about OS. Should this file be called FfmpegCapturer or something more generic to hide the Ffmpeg detail?

src/main/java/org/jitsi/jibri/capture/ffmpeg/executor/FfmpegExecutor.kt
```diff
@@ -0,0 +1,43 @@
...
+    val streamingBufSize: Int = streamingMaxBitrate * 2,
+    // The range of the CRF scale is 0–51, where 0 is lossless,
+    // 23 is the default, and 51 is worst quality possible. A lower value
+    // generally leads to higher quality, and a subjectively sane range is
+    // 17–28. Consider 17 or 18 to be visually lossless or nearly so;
+    // it should look the same or nearly the same as the input but it
+    // isn't technically lossless.
+    // https://trac.ffmpeg.org/wiki/Encode/H.264#crf
+    val h264ConstantRateFactor: Int = 25,
+    val gopSize: Int = framerate * 2
```
virtuacoplenny: Are these defaults or hardcoded values? If hardcoded, do you ever foresee that being an issue that's tricky to overcome?

src/test/java/org/jitsi/jibri/util/XmppUtilsKtTest.kt
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Test
+    fun `test roomjid with subdomain url conversion`() {
+        val expected = CallUrlInfo("https://$baseDomain", "roomName")
+        val jid = JidCreate.entityBareFrom("${expected.callName}@mucdomain.$baseDomain")
+
+        expect(expected) { getCallUrlInfoFromJid(jid, "mucdomain.", "brian.jitsi.net") }
+    }
+}
```
virtuacoplenny: Didn't you say the kotlin linter yells about not having a new line at the end of the file?

src/test/java/org/jitsi/jibri/CallUrlInfoTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    @Test
+    fun `test creation`() {
+        val info = CallUrlInfo("baseUrl", "callName")
+        assertEquals("baseUrl", info.baseUrl)
+        assertEquals("callName", info.callName)
+        assertEquals("baseUrl/callName", info.callUrl)
+    }
+
+    @Test
+    fun `test equals() and hashCode()`() {
```
virtuacoplenny: You weighed having two tests vs one test and decided it was worth it to keep the tests together?

virtuacoplenny: Also, is overriding "==" a kotlin thing or standard in some version of java?

doc/apis.md
```diff
@@ -0,0 +1,15 @@
+# Jibri APIs
+Jibri has two types of APIs:
+1. One is for 'external' use and is used to control stopping and starting of Jibri services, as well as querying Jibri's health.
```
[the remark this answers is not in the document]
bbaldino: yeah this is left as a matter for deployment to solve.

doc/internal_http_api.md
```diff
@@ -0,0 +1,13 @@
...
+##### URL
+`/jibri/api/internal/v1.0/notifyConfigChanged`
+##### Method
+`POST`
+##### URL Params
+None
+##### Data Params
+None
+##### Response
+If Jibri is currently idle, no respose will be sent as Jibri will shutdown immediately.  If Jibri is currently busy, it will respond with a `200`
```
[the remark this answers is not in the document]
bbaldino: this request tells jibri to exit as soon as it's idle, if it's currently idle it will shutdown before the code reaches the point to send the response. if i want to send the response first, i'd have to launch the 'shutdown next time idle' task in a thread or something, which i could do.  but the old version used a signal (which had no notion of a response at all) so i thought this behavior felt ok.

doc/http_api.md
```diff
@@ -0,0 +1,75 @@
...
+```
+{
+    "busy": Boolean // true if Jibri is currently busy, false otherwise
+}
+```
+This call should always respond with a `200` and the status encoded in the response body.  The lack of any response would represent an error of some sort
+
+
+##### URL
+`/jibri/api/v1.0/startService`
```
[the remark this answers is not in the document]
bbaldino: this is copied from old jibri, but allowing for the ability to support a minor version change doesn't seem like the worst thing.

resources/finalize_recording.sh
bbaldino: so that the code can assume the file will exist and it's clearer what a user needs to do to write their own

src/main/java/org/jitsi/jibri/CallUrlInfo.kt
bbaldino: it's an inherited member variable

src/main/java/org/jitsi/jibri/JibriManager.kt
bbaldino: this is a private function and it's only called from methods which are all marked `@Synchronized`

src/main/java/org/jitsi/jibri/JibriManager.kt
bbaldino: the "busy" state is based on the value of `currentActiveService` which won't get assigned until after this line, so if we return here we never made it to busy.

src/main/java/org/jitsi/jibri/api/http/HttpApi.kt
bbaldino: this is an enum and those are the only values, so it can't be anything else.

src/main/java/org/jitsi/jibri/api/http/HttpApi.kt
bbaldino: yeah have to do all those

bbaldino:
Thanks for taking a look @bgrozev 
> I haven't read this carefully (I will try to take a closer look later), but a couple of observations:
> Since we're now using Kotlin, I think we should agree on some code conventions and follow them everywhere. 

Agree, all the files here follow the Kotlin style guide and there is a task in `mvn verify` which enforces them.  

> I would suggest to adopt the java convention wherever appropriate, as kotlin and java code are going to appear together often. What do we do about:

I think for the most part this will be true.

> licence headers

Yes these should be here, I just forgot to add them to my intellij template for .kt files

> line length

Agree

> imports (by-class vs *)

The Kotlin style guide says to avoid * imports so I looked into this (* vs individual) and after doing so I believe individual ones are better.  So between that and the style guide itself recommending it I went that way.  Perhaps something to discuss.
> braces

Kotlin style puts this on the same line.  I suspect one reason is that Kotlin has a nice syntax for passing a lambda function.

> My other questions is about the xmpp/muc code. Are we still planning to have a common base in jicoco that we can reuse in both jitsi-videobridge and jibri?

Yes, the muc code has a comment here that it will be moved.  Its package structure in jibri was set up in a way that simulates (mostly) how it will look once it's moved.

src/main/java/org/jitsi/jibri/api/http/internal/InternalHttpApi.kt
```diff
@@ -0,0 +1,28 @@
...
+    @POST
+    @Path("notifyConfigChanged")
+    fun reloadConfig(): Response {
+        logger.debug("Config file changed")
+        configChangedHandler()
+        // If Jibri is currently idle, then we'll exit in the context
+        // of the above call, so this response will never get sent.
+        // Since the old Jibri used a signal (which has no concept of
+        // a response), I'm assuming it will be ok if we don't return one
+        // in some instances here.
```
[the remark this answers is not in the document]
bbaldino: i know that the old code had no notion of this, so it certainly won't regress anything.  if we really need the response, i can launch the shutdown task in a thread and send the response.

src/main/java/org/jitsi/jibri/api/xmpp/JibriSyncIqRequestHandler.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.jivesoftware.smack.packet.IQ
+import org.jivesoftware.smack.packet.XMPPError
+
+abstract class JibriSyncIqRequestHandler : AbstractIqRequestHandler(
+        JibriIq.ELEMENT_NAME,
+        JibriIq.NAMESPACE,
+        IQ.Type.set,
+        IQRequestHandler.Mode.sync) {
+    override fun handleIQRequest(iq: IQ): IQ {
+        if (iq !is JibriIq) {
```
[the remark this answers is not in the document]
bbaldino: yeah i think i like that better

src/main/java/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,214 @@
...
+     * connection, we'll listen for incoming [JibriIq] messages and handle them appropriately.  Join the MUC on
+     * each connection and send an initial [JibriStatusPacketExt] presence.
+     */
+    fun start() {
+        JibriStatusPacketExt.registerExtensionProvider()
+        ProviderManager.addIQProvider(
+            JibriIq.ELEMENT_NAME, JibriIq.NAMESPACE, JibriIqProvider()
+        )
+
+        // Join all the mucs we've been told to
```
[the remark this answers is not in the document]
bbaldino: the goal is to use camel case in code and proper caps in comments/docs, i'm sure i'm missing a few.  will fix this one.

src/main/java/org/jitsi/jibri/api/xmpp/XmppApi.kt
bbaldino: config is validated when it's parsed (in terms of the proper fields being present).  there may be some more semantic validating we could do on the field values themselves...maybe worth doing some negative testing to make sure things are handled ok.

src/main/java/org/jitsi/jibri/api/xmpp/XmppApi.kt
bbaldino: yeah i'll probably play with this a bit.

src/main/java/org/jitsi/jibri/api/xmpp/XmppApi.kt
bbaldino: unfortunately there aren't many constructors in JibriIq.  I already added a helper function for creating an iq response, i think it would be nice to add a helper for this case too.

src/main/java/org/jitsi/jibri/capture/ffmpeg/FfmpegCapturer.kt
bbaldino: well `FfmpegCapturer` basically is that factory, this is where this logic is being hidden (it doesn't do much else).  And Ffmpeg in the name is intentional, this isn't the generic part (that's just `Capturer`) this is the specific type that denotes it as an _ffmpeg_ Capturer

src/main/java/org/jitsi/jibri/capture/ffmpeg/executor/FfmpegExecutor.kt
bbaldino: these are hard-coded.  the old one had these hard-coded, so i didn't feel the need to handle being able to dynamically set these values until we come across that use case.

src/test/java/org/jitsi/jibri/util/XmppUtilsKtTest.kt
bbaldino: i don't think the linter is running on the test files actually.  i don't have it configured to skip them, so maybe it's a default setting. will have to look into it.

src/test/java/org/jitsi/jibri/CallUrlInfoTest.kt
bbaldino: i did these as one because equals/hashcode are defined based on one another (hashcode must be equal if == returns true).  overriding == is in java as well.

README.md
```diff
@@ -1,246 +1,112 @@
...
 
 
 # Installing Jibri
 
-### Installation Notes
-
-* Jibri was built on ubuntu 16 (Xenial), and has been tested with the pre-built kernel and extra kernel modules ( linux-image-extra-virtual package).  Any other distribution or kernel configuration MAY work but have not been tested.
-
+### Installation notes
+* Jibri was built on ubuntu 16 (Xenial), and has been tested with the pre-built kernel and extra kernel modules ( linux-image-extra-virtual package). Any other distribution or kernel configuration MAY work but has not been tested.
```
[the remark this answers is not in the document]
bbaldino: fixed

README.md
```diff
@@ -1,246 +1,112 @@
...
-
-### Jibri directory
-
-* Set up the jibri home directory (for example /home/jibri)
-
-* Check out the jibri source code: `git clone https://github.com/jitsi/jibri.git`
-
-* Copy the two main directories into place.  In this example install, the ```jibri-xmpp-client``` and ```scripts``` directories should be copied to /home/jibri/ so that they are siblings.  Many scripts referenced in the jibri daemon currently assume the scripts directory to be available from ../scripts/
+### Google chrome stable & Chromedriver
+The latest google chrome stable build should be used. It may be able to be installed direclty via apt, but the manual instructions for installing it are as follows:
```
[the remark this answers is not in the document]
bbaldino: fixed

README.md
```diff
@@ -1,246 +1,112 @@
...
+unzip ~/chromedriver_linux64.zip -d ~/
+rm ~/chromedriver_linux64.zip
+sudo mv -f ~/chromedriver /usr/local/bin/chromedriver
+sudo chown root:root /usr/local/bin/chromedriver
+sudo chmod 0755 /usr/local/bin/chromedriver
+```
 
-  `cp -a jibri/jibri-xmpp-client /home/jibri`
-  `cp -a jibri/scripts /home/jibri`
+### Miscelaneous required tools
```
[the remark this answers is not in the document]
bbaldino: fixed

README.md
```diff
@@ -1,246 +1,112 @@
...
-  - `pip3 install setuptools`
-  - `pip3 install -r jibri/requirements.txt`
+* Edit the config.json file (installed to /etc/jitsi/jibri/config.json by default) appropriately.
 
+### Logging
+By default, Jibri logs to `/var/log/jitsi/jibri`.  If you don't install via the debian package, you'll need to make sure this directory exists (or change the location to which Jibri logs by editing the [log config](lib/logging.properties)
 
 ### Support daemon systemd configurations
+* Jibri expects XOrg and icewm to be running as daemons. To accomplish this, copy the example systemd files from the systemd directory into the appropriate place (/etc/systemd/system/ on Ubuntu 16).
+  * TODO: (where will the icewm/xorg systemd scripts be put?)
```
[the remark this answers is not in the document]
bbaldino: fixed

doc/debian_package.md
```diff
@@ -0,0 +1,15 @@
+# Building the Jibri debian package
+Building the debian package has been tested on Ubuntu Xenial.
+
+### Steps
+* Install the prerequisites:
+  * `sudo apt install git maven openjdk-8-jdk`
+  * (`openjdk-8-jdk` specifically is not necessarily required, any java 8 jdk will probably work)
+* Clone the repo:
+  * `git clone https://github.com/bbaldino/jibri-temp.git`
```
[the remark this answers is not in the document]
bbaldino: i was't sure what repo we were gonna use.  fixed.

src/main/java/org/jitsi/jibri/capture/ffmpeg/executor/impl/AbstractFfmpegExecutor.kt
```diff
@@ -0,0 +1,120 @@
...
+            if (!parsedOutputLine.containsKey("frame")) {
+                logger.error("Ffmpeg is running but doesn't appear to be encoding.  " +
+                        "Its most recent output line was $ffmpegOutput")
+                return false
+            } else {
+                logger.debug("Ffmpeg appears healthy: $parsedOutputLine")
+                return true
+            }
+        }
+        return false
```
virtuacoplenny: Okay, I'm confused on this kotlin syntax. `currentFfmpegProc?.let` is essentially a `if currentFfmpegProc { // do logic }`? (It almost looked to to me you were passing in some function. I don't see you using this .let {} syntax often.)

src/main/java/org/jitsi/jibri/capture/ffmpeg/executor/impl/AbstractFfmpegExecutor.kt
```diff
@@ -0,0 +1,120 @@
...
+            logger.info("Sending SIGINT to ffmpeg proc $pid")
+            Runtime.getRuntime().exec("kill -s SIGINT $pid")
+        } ?: run {
+            logger.info("Ffmpeg had already exited")
+        }
+        currentFfmpegProc?.waitFor(10, TimeUnit.SECONDS)
+        currentFfmpegProc?.isAlive.let {
+            // This isn't great, as killing ffmpeg this way will corrupt
+            // the entire recording (from what I've seen)
+            currentFfmpegProc?.destroyForcibly()
```
virtuacoplenny: Then what is the alternative to forcibly killing? Is 10 seconds an informed number?

src/main/java/org/jitsi/jibri/capture/ffmpeg/executor/impl/LinuxFfmpegExecutor.kt
```diff
@@ -0,0 +1,30 @@
...
+            -f alsa
+            -thread_queue_size ${ffmpegExecutorParams.queueSize}
+            -i hw:0,1,0
+            -acodec aac -strict -2 -ar 44100
+            -c:v libx264 -preset ${ffmpegExecutorParams.videoEncodePreset}
+            ${sink.options} -pix_fmt yuv420p -r ${ffmpegExecutorParams.framerate}
+            -crf ${ffmpegExecutorParams.h264ConstantRateFactor}
+            -g ${ffmpegExecutorParams.gopSize} -tune zerolatency
+            -f ${sink.format} ${sink.path}
+        """.trimIndent().replace("\n", " ")
```
virtuacoplenny: I really like this formatting. It makes it read like a map. I don't understand any of it but it looks like I should be able to.

src/main/java/org/jitsi/jibri/capture/ffmpeg/executor/impl/MacFfmpegExecutor.kt
```diff
@@ -0,0 +1,26 @@
...
+            -video_size ${ffmpegExecutorParams.resolution}
+            -i 0:0
+            -vsync 2
+            -acodec aac -strict -2 -ar 44100
+            -c:v libx264 -preset ${ffmpegExecutorParams.videoEncodePreset}
+            ${sink.options} -pix_fmt yuv420p -crf ${ffmpegExecutorParams.h264ConstantRateFactor}
+            -g ${ffmpegExecutorParams.gopSize} -tune zerolatency
+            -f ${sink.format} ${sink.path}
+        """.trimIndent().replace("\n", " ")
+    }
```
virtuacoplenny: I wonder if it would have been better to pass this into the constructor instead of creating a separate class, thereby passing around a "algo" instead of keeping it tied to an implementation.

src/main/java/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -0,0 +1,98 @@
...
+                return true;
+            } catch (e) {
+                return e.message;
+            }
+        """.trimMargin())
+        return when (result) {
+            is Boolean -> result
+            else -> false
+        }
+    }
```
virtuacoplenny: I recall there was talk of re-using the speaker stats stuff. What was insufficient with them?

src/main/java/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -0,0 +1,147 @@
...
+     */
+    private fun addParticipantTracker() {
+        CallPage(chromeDriver).injectParticipantTrackerScript(chromeDriver)
+    }
+
+    /**
+     * Join a a web call with Selenium
+     */
+    fun joinCall(callName: String): Boolean {
+        HomePage(chromeDriver).visit(CallUrlInfo(baseUrl, ""))
```
virtuacoplenny: I think how this reads.

src/main/java/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -0,0 +1,98 @@
...
+        } else {
+            return listOf()
+        }
+    }
+
+    fun leave() {
+        hangUpButton?.click()
+        //TODO: wait for call to be fully exited?
+    }
+}
```
virtuacoplenny: Clean work.

src/main/java/org/jitsi/jibri/service/JibriService.kt
```diff
@@ -0,0 +1,27 @@
...
+}
+
+typealias JibriServiceStatusHandler = (JibriServiceStatus) -> Unit
+
+/**
+ * Interface implemented by all implemented [JibriService]s.  A [JibriService]
+ * is responsible for an entire feature of Jibri, such as recording a call
+ * to a file or streaming a call to a url.
+ */
+abstract class JibriService : StatusPublisher<JibriServiceStatus>() {
```
virtuacoplenny: interface*

src/main/java/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * @see [JibriService.start]
+     */
+    override fun start(): Boolean {
+        if (!jibriSelenium.joinCall(recordingOptions.callParams.callUrlInfo.callName)) {
+            logger.error("Selenium failed to join the call")
+            stop()
+            return false
```
virtuacoplenny: Without much context, this seems like an "exceptional" case. When do you decide, in general, to throw an error versus return a status?

src/main/java/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -0,0 +1,158 @@
...
+            }
+            if (numRestarts == FFMPEG_RESTART_ATTEMPTS) {
+                logger.error("Giving up on restarting the capturer")
+                publishStatus(JibriServiceStatus.ERROR)
+                stop()
+            } else {
+                numRestarts++
+                // Re-create the sink here because we want a new filename
+                sink = FileSink(recordingOptions.recordingDirectory, recordingOptions.callParams.callUrlInfo.callName)
+                capturer.start(sink)
```
virtuacoplenny: Does this need some kind of backoff or wait time or is immediate retry sufficient?

src/main/java/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -0,0 +1,158 @@
...
+        capturer.start(sink)
+        var numRestarts = 0
+        val processMonitor = ProcessMonitor(capturer) { exitCode ->
+            if (exitCode != null) {
+                logger.error("Capturer process is no longer healthy.  It exited with code $exitCode")
+            } else {
+                logger.error("Capturer process is no longer healthy but it is still running, stopping it now")
+                capturer.stop()
+            }
+            if (numRestarts == FFMPEG_RESTART_ATTEMPTS) {
```
virtuacoplenny: I've generally found it safer to say >= in cases like this. Logically I don't think it's that different to read but covers more group.

src/main/java/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -0,0 +1,158 @@
...
+            val finalizeProc = Runtime.getRuntime()
+                .exec("${recordingOptions.finalizeScriptPath} ${recordingOptions.recordingDirectory}")
+            finalizeProc.waitFor()
+            logger.info("Recording finalize script finished with exit " +
+                    "value: ${finalizeProc.exitValue()}")
+        } catch (e: IOException) {
+            logger.error("Failed to run finalize script: $e")
+        }
+    }
+}
```
virtuacoplenny: I really like the JibriService and in general how easy it is to use the two services.

src/main/java/org/jitsi/jibri/util/ProcessMonitor.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * Run a check of [processToMonitor] to check if it is alive.  If it is
+     * dead, [processUnhealthyCallback] will be invoked with the exit code from
+     * [processToMonitor] (or null, if there isn't one)
+     */
+    override fun run() {
+        try {
+            if (!processToMonitor.isHealthy()) {
+                processUnhealthyCallback(processToMonitor.getExitCode())
```
virtuacoplenny: I'm wondering why limit the callback to unhealthy only. Is it cleaner to limit this to unhealthy callbacks only at this point?

src/main/java/org/jitsi/jibri/capture/ffmpeg/executor/impl/AbstractFfmpegExecutor.kt
bbaldino: yup, that's it.  main reason for doing it this way is as follows:

if you do
```
if (currentFfmpegProc != null) {
   // logic
}
```
you don't actually get a guarantee that currentFfmpegProc will not be null inside the if, as another thread could have come along and changed it.  so you'd do something like
```
val currentFfmpegProcCopy = currentFfmpegProc // currentFfmpegProcCopy is an immutable type
if (currentFfmpegProcCopy) {
   // logic
}
```
since we copied the reference into an immutable variable, we know it couldn't have been changed to null in between the if check and the block inside.  `let` is basically just shortcut code to do this.

src/main/java/org/jitsi/jibri/capture/ffmpeg/executor/impl/AbstractFfmpegExecutor.kt
bbaldino: i don't know of any alternative :( and 10 is not terribly significant, just seemed like a reasonable value.

src/main/java/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -0,0 +1,147 @@
...
+    private val logger = Logger.getLogger(this::class.qualifiedName)
+    private var chromeDriver: ChromeDriver
+    private val baseUrl: String = jibriSeleniumOptions.callParams.callUrlInfo.baseUrl
+    /**
+     * The options we'll add as url params
+     */
+    private val urlOptions = listOf(
+        "config.iAmRecorder=true",
+        "config.externalConnectUrl=null",
+        "interfaceConfig.APP_NAME=\"Jibri\""
```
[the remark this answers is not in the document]
bbaldino: these are used to make jibri not appear like a typical participant (be 'invisible')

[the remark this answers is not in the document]
bbaldino: i can't remember exactly what the externalconnecturl thing was, but i remember aaron explaining to me why it was necessary.

src/main/java/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -0,0 +1,147 @@
...
+                "--enabled",
+                "--enable-logging",
+                "--vmodule=*=3",
+                "--disable-infobars",
+                "--alsa-output-device=plug:amix"
+        )
+        val chromeDriverService = ChromeDriverService.Builder().withEnvironment(
+            mapOf("DISPLAY" to jibriSeleniumOptions.display)
+        ).build()
+        chromeDriver = ChromeDriver(chromeDriverService, chromeOptions)
```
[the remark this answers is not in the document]
bbaldino: what fake devices do you mean here?  jibri doesn't send anyway, so we don't actually want anything.

(in my branch with the sipgateway changes where we do need to send video, this changes to become more flexible)

src/main/java/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
bbaldino: i don't remember why, but i do know i looked at it and went this direction.

src/main/java/org/jitsi/jibri/service/JibriService.kt
bbaldino: fixed

bbaldino: actually nm, this can't be an interface because it inherits from `StatusPublisher`

src/main/java/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
bbaldino: nah no backoff needed here

src/main/java/org/jitsi/jibri/util/ProcessMonitor.kt
bbaldino: yeah right now things are only interested in unhealthy-ness.

src/main/kotlin/org/jitsi/jibri/api/http/HttpApi.kt
```diff
@@ -0,0 +1,108 @@
...
+        val result: StartServiceResult = when (serviceParams.sinkType) {
+            RecordingSinkType.FILE -> run {
+                jibriManager.startFileRecording(
+                    ServiceParams(usageTimeoutMinutes = 0),
+                    FileRecordingParams(serviceParams.callParams),
+                    environmentContext = null
+                )
+            }
+            RecordingSinkType.STREAM -> run {
+                val youTubeStreamKey = serviceParams.youTubeStreamKey ?: return@run StartServiceResult.ERROR
```
bgrozev: Nit: the val is unused, perhaps you want to use it on the line below instead of accessing serviceParams again?

src/main/kotlin/org/jitsi/jibri/api/http/internal/InternalHttpApi.kt
```diff
@@ -0,0 +1,68 @@
...
+     * it is idle (i.e. finish any currently running service). Returns a 200
+     * and schedules a shutdown for when it becomes idle.
+     */
+    @POST
+    @Path("gracefulShutdown")
+    fun gracefulShutdown(): Response {
+        logger.info("Jibri gracefully shutting down")
+        // Schedule firing the handler so we have a chance to send the successful
+        // response.
+        executor.schedule(1) {
```
bgrozev: This is just me trying to wrap my head around Kotlin things: gracefulShutdownHandler is already a `() -> Unit`, why do we wrap it in a second one? Wouldn't it be cleaner to do
`executor.schedule(1, action = gracefulShutdownHandler)`?

src/main/kotlin/org/jitsi/jibri/api/http/internal/InternalHttpApi.kt
```diff
@@ -0,0 +1,68 @@
...
+     * running and shutdown.  Returns a 200 and schedules a shutdown with a 1
+     * second delay.
+     */
+    @POST
+    @Path("shutdown")
+    fun shutdown(): Response {
+        logger.info("Jibri is forecfully shutting down")
+        // Schedule firing the handler so we have a chance to send the successful
+        // response.
+        executor.schedule(1) {
```
bgrozev: Same as above, and a typo in the log message

src/main/kotlin/org/jitsi/jibri/api/http/internal/InternalHttpApi.kt
```diff
@@ -0,0 +1,68 @@
...
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+import java.util.logging.Logger
+import javax.ws.rs.POST
+import javax.ws.rs.Path
+import javax.ws.rs.core.Response
+
+@Path("/jibri/api/internal/v1.0")
+class InternalHttpApi(
+    private val executor: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor(),
```
bgrozev: I see we have an easy way to have the threads named (used in XmppApi.kt). Is it intentionally not used here?

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,245 @@
...
+                try {
+                    val mucClient = MucClient(configBuilder.build())
+                    mucClient.addIqRequestHandler(object : JibriSyncIqRequestHandler() {
+                        override fun handleJibriIqRequest(jibriIq: JibriIq): IQ {
+                            return handleJibriIq(jibriIq, config, mucClient)
+                        }
+                    })
+                    jibriManager.addStatusHandler { status ->
+                        logger.info("XMPP API got jibri status $status, publishing presence to connection ${config.name}")
+                        val jibriStatus = JibriPresenceHelper.createPresence(
```
bgrozev: What do you think about renaming `jibriStatus` to `presence`? With the types being implicit I find it even more useful for the name to reflect the type.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,245 @@
...
+                        mucClient.sendStanza(jibriStatus)
+                    }
+                    mucClient.createOrJoinMuc(
+                        JidCreate.entityBareFrom("${config.controlMuc.roomName}@${config.controlMuc.domain}"),
+                        Resourcepart.from(config.controlMuc.nickname)
+                    )
+
+                    val jibriStatus = JibriPresenceHelper.createPresence(
+                        JibriStatusPacketExt.Status.IDLE,
+                        JidCreate.bareFrom("${config.controlMuc.roomName}@${config.controlMuc.domain}")
```
bgrozev: We create the same Jid three times in this block. Worth extracting into a local variable? I think we can use EntityBareJid everywhere.

bgrozev: Same note as above about the name `jibriStatus`

src/main/kotlin/org/jitsi/jibri/api/xmpp/JibriIqHelper.kt
```diff
@@ -0,0 +1,67 @@
...
+        fun create(from: Jid, type: IQ.Type, status: JibriIq.Status): JibriIq {
+            val jibriIq = JibriIq()
+            jibriIq.to = from
+            jibriIq.type = type
+            jibriIq.status = status
+
+            return jibriIq
+        }
+
+        fun create(from: Jid, type: IQ.Type): JibriIq {
```
bgrozev: Why do we have two methods instead of using default parameter values? The type parameter could also default to Type.set (which would allow XmppApi not not override it)

src/main/kotlin/org/jitsi/jibri/api/http/HttpApi.kt
bbaldino: yup, thanks.  annoying that linter didn't catch this.  i'll have to look into that.

src/main/kotlin/org/jitsi/jibri/api/http/internal/InternalHttpApi.kt
bbaldino: yup, totally right.  i was probably just use to throwing lambdas everywhere.

src/main/kotlin/org/jitsi/jibri/api/http/internal/InternalHttpApi.kt
bbaldino: i added this midway through and meant to go back and clean them all up to use it but forgot looks like.  thanks.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
bbaldino: ya good call

src/main/kotlin/org/jitsi/jibri/api/xmpp/JibriIqHelper.kt
bbaldino: yeah i think i was in the mindset of these being moved to `JibriIq` so i was writing them a bit java-style.  i'll fix it.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,245 @@
...
+                // to notify the caller who invoked the service of its status, so we'll listen
+                // for the service's status while it's running and this method will be invoked
+                // if it changes
+                val serviceStatusHandler: JibriServiceStatusHandler = { serviceStatus ->
+                    when (serviceStatus) {
+                        JibriServiceStatus.ERROR -> {
+                            logger.info("Current service had an error")
+                            val errorIq = JibriIqHelper.create(startJibriIq.from, IQ.Type.set, JibriIq.Status.FAILED)
+                            logger.info("Sending error iq ${errorIq.toXML()}")
+                            mucClient.sendStanza(errorIq)
```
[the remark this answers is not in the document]
bbaldino: yeah that's a good catch.  this is yet another place where i've had a need to basically say "schedule a task as soon as possible, but at least AFTER the current context finishes" (like `setTimeout({...}, 0)` in javascript).  i'll think about this.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,245 @@
...
+            } finally {
+                logger.info("Sending start service response iq: ${resultIq.toXML()}")
+                mucClient.sendStanza(resultIq)
+            }
+        }
+        // Immediately respond that the request is pending
+        val initialResponse = JibriIqHelper.createResult(startJibriIq)
+        initialResponse.status = JibriIq.Status.PENDING
+        logger.info("Sending 'pending' response to start IQ")
+        return initialResponse
```
bgrozev: If we add the status as a parameter to `JibriIqHelper.createResult` we can simplify to `return JibriIqHelper.createResult(...)`

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -0,0 +1,245 @@
...
+                        )
+                        mucClient.sendStanza(jibriStatus)
+                    }
+                    mucClient.createOrJoinMuc(
+                        JidCreate.entityBareFrom("${config.controlMuc.roomName}@${config.controlMuc.domain}"),
+                        Resourcepart.from(config.controlMuc.nickname)
+                    )
+
+                    val jibriStatus = JibriPresenceHelper.createPresence(
+                        JibriStatusPacketExt.Status.IDLE,
```
bgrozev: Perhaps import JibriIq.Status to shorten all the references a little bit?

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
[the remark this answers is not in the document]
bgrozev: Theoretically, XMPP is async, so the `result` and subsequent `set` IQs could arrive at jicofo out of order.

src/main/kotlin/org/jitsi/jibri/util/extensions/SchedulerExecutorServiceExts.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+
+/**
+ * A version of [ScheduledExecutorService.scheduleAtFixedRate] that takes
+ * a lambda instead of requiring a [Runnable] (and moves it to the last
+ * argument)
+ */
+fun ScheduledExecutorService.scheduleAtFixedRate(
+    delay: Long = 0,
```
bgrozev: Since we a redeviating from Java's argument order anyway, how about moving `delay` to the third place (since it has a default value)? We can then removes the named arguments from it's usage:
`scheduleAtFixedRate(period = 15, unit = TimeUnit.SECONDS) `

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
bbaldino: i agree they get a bit long, but with all the various statuses i worry it might get more confusing to not have them be fully qualified.

src/main/kotlin/org/jitsi/jibri/util/extensions/SchedulerExecutorServiceExts.kt
bbaldino: this actually exists to move the function to the last argument so the
```
scheduleAtFixedRate(delay, unit) {
 // func
}
```
syntax can be used, if the function isn't the last one we can't use the {} syntax :/

src/main/kotlin/org/jitsi/jibri/util/testHelpers/Eventually.kt
```diff
@@ -0,0 +1,43 @@
+/*
```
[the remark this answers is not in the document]
bbaldino: i originally had it in the tests, but i moved it out for some reason.  maybe it was just because i thought it looked a little confusing nestled amongst the tests in its own directory (where normally the test dirs exactly mirror the real test files).  but that doesn't seem like a big deal, so moved it back to the test dir.

src/main/kotlin/org/jitsi/jibri/util/extensions/SchedulerExecutorServiceExts.kt
bgrozev: My suggestion is to move `delay` (which has a default value) after `period` and `unit`:
```
fun ScheduledExecutorService.scheduleAtFixedRate(
    period: Long,
    unit: TimeUnit,
    delay: Long = 0,
    action: () -> Unit
):
```

This allows us to call with `scheduleAtFixedRate(15, TimeUnit.SECONDS)` instead of having to specify the argument names: `scheduleAtFixedRate(period = 15, unit = TimeUnit.SECONDS)`

src/main/kotlin/org/jitsi/jibri/service/impl/StreamingJibriService.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    init {
+        sink = StreamSink(
+            url = "$YOUTUBE_URL/${streamingOptions.youTubeStreamKey}",
+            streamingMaxBitrate = STREAMING_MAX_BITRATE,
+            streamingBufSize = 2 * STREAMING_MAX_BITRATE
+        )
+
+        // Bubble up jibriSelenium's status
+        jibriSelenium.addStatusHandler {
```
bgrozev: `jibriSelenium.addStatusHandler(this::publishStatus)`?

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * The [Logger] for this class
+     */
+    private val logger = Logger.getLogger(this::class.qualifiedName)
+    /**
+     * Used for the selenium interaction
+     */
+    private val jibriSelenium = JibriSelenium(
+        JibriSeleniumOptions(recordingOptions.callParams),
+        Executors.newSingleThreadScheduledExecutor(NameableThreadFactory("JibriSelenium"))
```
bgrozev: Perhaps we can move this executor instantiation to a default value in JibriSelenium (to keep the string reference local)?
I'm curious why for the streaming service JibriSelenium shares the service's executor, but for the file recording it has its own?

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -0,0 +1,174 @@
...
+        Executors.newSingleThreadScheduledExecutor(NameableThreadFactory("FileRecordingJibriService"))
+    /**
+     * The handle to the scheduled process monitor task, which we use to
+     * cancel the task
+     */
+    private var processMonitorTask: ScheduledFuture<*>? = null
+
+    init {
+        sink = FileSink(recordingOptions.recordingDirectory, recordingOptions.callParams.callUrlInfo.callName)
+        jibriSelenium.addStatusHandler {
```
bgrozev: `jibriSelenium.addStatusHandler(this::publishStatus)`?

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -0,0 +1,174 @@
...
+            publishStatus(it)
+        }
+    }
+
+    override fun start(): Boolean {
```
bgrozev: I wish we could avoid the duplication between `FileRecordingJibriService#start` and `StreamingJibriService#start`. They are very similar (and not trivial)

bgrozev: Same goes for `stop`. They both have a `capturer` and a `processMonitorTask`. Perhaps we can move these to super?

src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/util/OutputParser.kt
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+package org.jitsi.jibri.capture.ffmpeg.util
+
+import org.jitsi.jibri.util.decimal
+import org.jitsi.jibri.util.oneOrMoreNonSpaces
+import org.jitsi.jibri.util.zeroOrMoreSpaces
+import java.util.regex.Pattern
+
+const val ENCODING_KEY = "frame"
```
bgrozev: Perhaps some docs on what these are and how they are used?

If this is just a grep for "frame" perhaps that's a little too wide and could match lines that you didn't intend to?

src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/executor/FfmpegExecutor.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.jitsi.jibri.sink.Sink
+import org.jitsi.jibri.util.MonitorableProcess
+
+/**
+ * Paramaters which will be passed to ffmpeg
+ */
+data class FfmpegExecutorParams(
+    val resolution: String = "1280x720",
+    val framerate: Int = 30,
+    val videoEncodePreset: String = "veryfast",
```
bgrozev: Just curious: have we been using "veryfast" all along?

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -0,0 +1,161 @@
...
+     * Jibri is active
+     */
+    private fun addParticipantTracker() {
+        CallPage(chromeDriver).injectParticipantTrackerScript(chromeDriver)
+    }
+
+    /**
+     * Join a a web call with Selenium
+     */
+    fun joinCall(callName: String): Boolean {
```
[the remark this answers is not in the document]
bbaldino: you're right in practice but that feels a bit odd to me.  i think what actually would feel better would be not saving the `baseUrl` at all earlier, and just taking the whole `callUrlInfo` as a param here.

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
bbaldino: yeah, this stuff is a bit inconsistent.  executors were moved to be params in places as unit tests were written so that a mock could be passed in.  in this place i think i just happened to optimize a bit by having the service shared.  i need a good sweep on all the usages here to make them consistent.

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
bbaldino: yeah totally agree.  i tried to combine them earlier on but it was a bit awkward and because of some of the differences still didn't feel great.  i want to take another look at it now; especially with adding the sipgw stuff there is more duplication so hoping can find a way to reuse some code.

src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/executor/FfmpegExecutor.kt
bbaldino: yeah

src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/util/OutputParser.kt
bbaldino: yeah i'll add some docs here.  so far there hasn't been an issue with regards to catching something we didn't...but this is also used as a key from a string parsed from a `key=value` substring, so really this will only be matched against a `frame=XXX` type string.

src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/executor/impl/AbstractFfmpegExecutor.kt
```diff
@@ -0,0 +1,150 @@
...
+        ffmpegOutputLogger.addHandler(FfmpegFileHandler())
+    }
+
+    /**
+     * Get the shell command to use to launch ffmpeg
+     */
+    protected abstract fun getFfmpegCommand(ffmpegExecutorParams: FfmpegExecutorParams, sink: Sink): String
+
+    override fun launchFfmpeg(ffmpegExecutorParams: FfmpegExecutorParams, sink: Sink): Boolean {
+        processBuilder.command(getFfmpegCommand(ffmpegExecutorParams, sink).split(" "))
```
bgrozev: Probably not worth worrying about at this point, but I think this will misbehave if the path contains spaces.

bbaldino: i actually have been thinking about this situation, because in the sipgw case the arguments are formatted differently and this method doesn't work well, so going to try and see if i can come up with a better way

src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/executor/impl/AbstractFfmpegExecutor.kt
```diff
@@ -0,0 +1,150 @@
...
+        }
+        return false
+    }
+
+    override fun stopFfmpeg() {
+        executor.shutdown()
+        currentFfmpegProc?.let {
+            val pid = pid(it)
+            logger.info("Sending SIGINT to ffmpeg proc $pid")
+            Runtime.getRuntime().exec("kill -s SIGINT $pid")
```
bgrozev: Should we also send a SIGTERM?

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -0,0 +1,161 @@
...
+        emptyCallTask = executor.scheduleAtFixedRate(period = 15, unit = TimeUnit.SECONDS) {
+            try {
+                // >1 since the count will include jibri itself
+                if (CallPage(chromeDriver).getNumParticipants(chromeDriver) > 1) {
+                    numTimesEmpty = 0
+                } else {
+                    numTimesEmpty++
+                }
+                if (numTimesEmpty >= 2) {
+                    logger.info("Jibri has been in a lonely call for 30 seconds, marking as finished")
```
bgrozev: Since the task was submitted without a delay, I think this will actually trigger after 15 seconds, not 30.

src/test/kotlin/org/jitsi/jibri/CallUrlInfoTest.kt
```diff
@@ -0,0 +1,96 @@
...
+            should("assign the fields correctly") {
+                info.baseUrl shouldBe "baseUrl"
+                info.callName shouldBe "callName"
+                info.callUrl shouldBe "baseUrl/callName"
+            }
+        }
+        "a nullable CallUrlInfo instance" {
+            should("not equal null") {
+                val nullableInfo: CallUrlInfo? = CallUrlInfo("baseUrl", "callName")
+                (nullableInfo == null) shouldBe false
```
bgrozev: `nullableInfo shouldNotBe null`?

src/test/kotlin/org/jitsi/jibri/api/http/HttpApiTest.kt
```diff
@@ -0,0 +1,158 @@
...
+                    environmentContext = EnvironmentContext("meet.jit.si")
+                )
+                whenever(jibriManager.healthCheck())
+                    .thenReturn(expectedHealth)
+                should("return the right json body and a status of 200") {
+                    val res = jerseyTest.target("/jibri/api/v1.0/health").request()
+                        .get()
+                    res.status shouldBe 200
+                    val json = res.readEntity(String::class.java)
+                    // The json should not include the 'environmentContext' field at all, since it
```
bgrozev: Extra comment?

src/test/kotlin/org/jitsi/jibri/api/http/HttpApiTest.kt
```diff
@@ -0,0 +1,158 @@
...
+                    sinkType = RecordingSinkType.FILE
+                )
+                val json = jacksonObjectMapper().writeValueAsString(startServiceRequest)
+                should("return a 200") {
+                    val res = jerseyTest
+                        .target("/jibri/api/v1.0/startService")
+                        .request()
+                        .post(Entity.json(json))
+                    res.status shouldBe 200
+                    println(res.readEntity(String::class.java))
```
bgrozev: Intentional print?

src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/executor/impl/AbstractFfmpegExecutor.kt
bbaldino: i cleaned this up in the sip gateway branch

PR #84, opened 2018-04-23
Add instructions on how to install jibri via the public repository to README

aaronkvanmeerten:
These instructions should be more complete with the repo details included

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,26 @@ sudo chmod 0755 /usr/local/bin/chromedriver
 ### Miscellaneous required tools
 See the debian [control file](resources/debian-package-jibri/DEBIAN/control) for the dependencies that are required.
 
+### Jitsi Debian Repository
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,26 @@ sudo chmod 0755 /usr/local/bin/chromedriver
 ### Miscellaneous required tools
 See the debian [control file](resources/debian-package-jibri/DEBIAN/control) for the dependencies that are required.
 
+### Jitsi Debian Repository
+The latest Jibri packages can be found in the unstable repository on downloads.jitsi.org (https://jitsi.org/downloads/).
```
bbaldino: i don't see anything about jibri at this url?

bbaldino: i don't think it's worth mentioning this https url since there's nothing about jibri here.  maybe just changing this to something like "To install the Jibri debian package, do the following steps:"

PR #86, opened 2018-04-24
Participant metadata

bbaldino:


src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -203,6 +211,16 @@ class JibriSelenium(
...
+        File("/tmp/browser.out").printWriter().use { out ->
+            chromeDriver.manage().logs().availableLogTypes.forEach { logType ->
+                val logEntries = chromeDriver.manage().logs().get(logType)
+                logger.info("Got ${logEntries.all.size} log entries")
+                out.println("========= TYPE=$logType ===========")
+                logEntries.all.forEach {
+                    out.println(it.toJson())
+                }
+            }
+        }
```
virtuacoplenny: Is the additional logging left in on purpose? If so, do you really only want LogType.DRIVER logs?

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -82,10 +82,24 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
...
+                console.log("There were " + Object.keys(existingMembers).length + " existing members");
+                for (let key in existingMembers) {
+                    const existingMember = existingMembers[key];
+                    if (existingMember.identity) {
+                        console.log("Member ", existingMember, " has identity, adding");
+                        window._jibriParticipants.push(existingMember.identity);
+                    } else {
+                        console.log("Member ", existingMember.jid, " has no identity, skipping");
+                    }
+                }
```
virtuacoplenny: Do you have to take into account any "hidden" members, like the focus?

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -82,10 +82,24 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
         val result = driver.executeScript("""
             try {
                 window._jibriParticipants = [];
+                const existingMembers = APP.conference._room.room.members || {};
+                console.log("There were " + Object.keys(existingMembers).length + " existing members");
+                for (let key in existingMembers) {
```
virtuacoplenny: To be cute...you already did Object.keys so you might as well store it and do a forEach/reduce/filter/map

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -82,10 +82,24 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
...
+                    }
+                }
                 APP.conference._room.room.addListener(
                     "xmpp.muc_member_joined",
                     (from, nick, role, hidden, statsid, status, identity) => {
-                        window._jibriParticipants.push(identity);
+                        console.log("Jibri got MUC_MEMBER_JOINED: ", from, identity);
+                        if (identity) {
+                            window._jibriParticipants.push(identity);
+                        }
```
virtuacoplenny: So the identity truthy check wasn't done before. What does the truthy check give you?

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
bbaldino: So I was going to yank this out before pushing this, but then thought they would be useful to have.  Then I was thinking about it more this morning and realized in order for them to be actually useful, they should write to more proper log files with log rotation and all that (otherwise they get overwritten every call).  So I'll decide whether to yank this for now or make it more official.

bbaldino: Also, I found DRIVER to have the most useful information (for whatever reason).

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
bbaldino: from what I found, those won't have 'identity' so we'll be covered by the check later.

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
bbaldino: though as much as i'd like to do the map/filter/etc i want the debug printing in the 'else' cases to know what we're filtering out so will just do a loop.

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
bbaldino: previously 'null' could end up in this list, which we don't want.

PR #88, opened 2018-04-25
Log finalize script output to jibri logs

bbaldino:
Doing this required some other changes to make sure we got all the logs correctly...I separated these by commit but if it's too confusing still I can do separate PRs, as the pre-work was pretty independent.  Because of that, let's not squash this when merging.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/util/LoggingUtils.kt
+++ b/src/main/kotlin/org/jitsi/jibri/util/LoggingUtils.kt
@@ -19,19 +19,32 @@ package org.jitsi.jibri.util
 import java.io.BufferedReader
 import java.io.InputStream
 import java.io.InputStreamReader
+import java.util.concurrent.Callable
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
+import java.util.concurrent.Future
 import java.util.logging.Logger
 
+/**
+ * A helper function to log a given [InputStream] via the given [Logger].
+ * A future is returned which will be completed when the end of the given
+ * stream is reached.
+ */
 fun logStream(
     stream: InputStream,
     logger: Logger,
     executor: ExecutorService = Executors.newSingleThreadExecutor(NameableThreadFactory("StreamLogger"))
-) {
-    executor.submit {
+): Future<*> {
+    return executor.submit(Callable<Boolean> {
         val reader = BufferedReader(InputStreamReader(stream))
-        while (true) {
-            logger.info(reader.readLine())
+        var eof = false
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/jitsi/jibri/util/LoggingUtils.kt
```diff
@@ -19,19 +19,32 @@ package org.jitsi.jibri.util
...
-) {
-    executor.submit {
+): Future<*> {
+    return executor.submit(Callable<Boolean> {
         val reader = BufferedReader(InputStreamReader(stream))
-        while (true) {
-            logger.info(reader.readLine())
+        var eof = false
+        while (!eof) {
+            reader.readLine()?.let {
```
bgrozev: This looks confusing to me, perhaps because I'm not used to Kotlin. Wouldn't something like this be easier to understand?
```
        while(true) {
            val line = reader.readLine()
            line?: return@Callable
            logger.info(line)
}
```

src/main/kotlin/org/jitsi/jibri/util/Tee.kt
```diff
@@ -59,9 +69,10 @@ class TeeLogic(inputStream: InputStream) {
      * starting from the time of its creation
      */
     fun addBranch(): InputStream {
-        val outputStream = PipedOutputStream()
-        branches.add(outputStream)
-        return PipedInputStream(outputStream)
+        PipedOutputStream().let {
```
bgrozev: Just curious: are these two equivalent? Is the purpose of using "let" here just to scope the PipedOutputStream instance?

src/main/kotlin/org/jitsi/jibri/util/LoggingUtils.kt
bbaldino: hmm, i think this would work...do you think it's better?
```
while (true) {
    val line = reader.readLine() ?: return@Callable true
    logger.info(line)
}
```

bgrozev: That's easier for me to understand (and I think getting rid of the extra var is good).

src/main/kotlin/org/jitsi/jibri/util/Tee.kt
bbaldino: yeah, it just saves declaring the variable basically.

actually i wonder if this would be better:
```
with(PipedOutputStream()) {
    branches.add(this)
    return PipedInputStream(this)
}
```

PR #102, opened 2018-05-08
fix an issue with restarting ffmpeg (without adding another bug this

bbaldino:
time)

was shutting down the executor, which means it will reject all future
tasks we try to submit to it.  since we use the same ffmpegexecutor
instance on restart, this was breaking restarting ffmpeg.  Also make
sure we let Tee run to its natural completion, so we get a proper
closing of all of its downstream branches.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/util/ProcessWrapper.kt
+++ b/src/main/kotlin/org/jitsi/jibri/util/ProcessWrapper.kt
@@ -88,8 +88,10 @@ abstract class ProcessWrapper<out StatusType>(
     fun waitFor(timeout: Long, unit: TimeUnit): Boolean = process.waitFor(timeout, unit)
 
     fun stop() {
-        tail.stop()
-        tee.stop()
+        // Note that we specifically do NOT call stop on tee or tail here
+        // because we want them to read everything available from the
+        // process' inputstream. Once it's done, they'll read
+        // the EOF and close things up correctly
         val pid = pid(process)
         Runtime.getRuntime().exec("kill -s SIGINT $pid")
     }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/jitsi/jibri/util/ProcessWrapper.kt
```diff
@@ -88,8 +88,10 @@ abstract class ProcessWrapper<out StatusType>(
     fun waitFor(timeout: Long, unit: TimeUnit): Boolean = process.waitFor(timeout, unit)
 
     fun stop() {
-        tail.stop()
-        tee.stop()
+        // Note that we specifically do NOT call stop on tee or tail here
+        // because we want them to read everything available from the
+        // process' inputstream. Once it's done, they'll read
+        // the EOF and close things up correctly
```
bbaldino: these lines were the cause of the bug that got introduced with the original version of this fix.  because shutdown wasn't called but we stopped these tasks, tee never read EOF which means it never closed the inputstreams that others were reading from, which means that when we try and wait for the `logStream` task to finish (the `processLoggerTask?.get()` above, which waits until it reads the last bit from the stream) we blocked forever.

src/main/kotlin/org/jitsi/jibri/util/ProcessWrapper.kt
```diff
@@ -88,8 +88,10 @@ abstract class ProcessWrapper<out StatusType>(
     fun waitFor(timeout: Long, unit: TimeUnit): Boolean = process.waitFor(timeout, unit)
 
     fun stop() {
-        tail.stop()
-        tee.stop()
```
bbaldino: these lines were the cause of the bug that got introduced with the original version of this fix. because shutdown wasn't called but we stopped these tasks, tee never read EOF which means it never closed the inputstreams that others were reading from, which means that when we try and wait for the logStream task to finish (the processLoggerTask?.get() above, which waits until it reads the last bit from the stream) we blocked forever.

PR #103, opened 2018-05-10
New recording flows

bbaldino:
This will need to update the jitsi-protocol-jabber dependency from [this PR](https://github.com/jitsi/jitsi/pull/475) so don't merge yet.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/api/http/HttpApi.kt
+++ b/src/main/kotlin/org/jitsi/jibri/api/http/HttpApi.kt
@@ -89,8 +89,8 @@ class HttpApi(private val jibriManager: JibriManager) {
                 // If it's a file recording, it must have the callLoginParams set
                 val callLoginParams = startServiceParams.callLoginParams ?: return@run StartServiceResult.ERROR
                 jibriManager.startFileRecording(
-                    ServiceParams("", usageTimeoutMinutes = 0),
-                    FileRecordingRequestParams(startServiceParams.callParams, callLoginParams),
+                    ServiceParams(usageTimeoutMinutes = 0),
+                    FileRecordingRequestParams(startServiceParams.callParams, "TODO: sessionID", callLoginParams),
                     environmentContext = null
                 )
             }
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/jitsi/jibri/api/http/HttpApi.kt
```diff
@@ -90,7 +90,7 @@ class HttpApi(private val jibriManager: JibriManager) {
                 val callLoginParams = startServiceParams.callLoginParams ?: return@run StartServiceResult.ERROR
                 jibriManager.startFileRecording(
                     ServiceParams(usageTimeoutMinutes = 0),
-                    FileRecordingRequestParams(startServiceParams.callParams, callLoginParams),
+                    FileRecordingRequestParams(startServiceParams.callParams, "TODO: sessionID", callLoginParams),
```
bbaldino: the HTTP API is not used in production right now, so these TODOs are ok.  I have a TODO list item to come back and get this to parity with the XmppApi, but really right now it's just useful for easier testing of Jibri.

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -143,6 +143,19 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
...
+                return e.message;
+            }
+            """.trimMargin())
+        return when (result) {
+            is String -> false
+            else -> true
+        }
+    }
+
+    fun sendPresence(driver: RemoteWebDriver): Boolean {
```
bgrozev: What does this do?

bbaldino: this makes a call in the web client to force the sending of its presence, so that we can propagate out the fields we added (session id, mode, etc.)

bbaldino:
tests are failing because it isn't using the right jitsi branch, will worry about that later.

Issue #104, opened 2018-05-11
Few installation tips and disambiguation for AWS EC2 users


realnumber:
Hi, I've managed to install Jitsi-meet and Jibri on a single EC2 instance and learned something that I wish I knew a few days ago so I thought I'd be nice to share:

1. [Installing ALSA loopback](https://github.com/jitsi/jibri/blob/master/README.md#alsa-and-loopback-device) module on default kernel fails so I followed the instructions [here](https://stackoverflow.com/questions/49538514/sound-driver-snd-aloop-kernel-module-setup-issue-with-aws-ec2-ubuntu-16-04-ins?answertab=active#tab-top) to make `modprobe snd-aloop` possible. That is, changing and updating the kernel boot settings from Linux 4.4.0-xxxx-aws to Linux 4.4.0-xxxx-generic and restarting the instance.

2. If you you run both jitsi-meet and jibri on a t2.micro instance, you're most likely encouter a memory overflow error like this as soon as you start streaming: `INFO: [50] ffmpeg.run() [x11grab @ 0xf4e940] Cannot get 3686432 bytes of shared memory: Cannot allocate memory.` In my case, switching the instance to a r4.large solved the issue.

3. (not only for AWS users) Despite what the [README.md](https://github.com/jitsi/jibri/blob/master/README.md#jitsi-meet) says, there is no "recording button" (I tried to find a button with this exact name). The "Go live now" is the button you are looking for :)

4. As for [user/group](https://github.com/jitsi/jibri/blob/master/README.md#user-group), I've installed and ran everything as root and it worked out perfectly.

By the way, the versions I used are 
```
Package: jitsi-meet
Version: 1.0.3035-1
Depends: jitsi-videobridge (= 1056-1), jicofo (= 1.0-409-1), jitsi-meet-web (= 1.0.2782-1), jitsi-meet-web-config (= 1.0.2782-1), jitsi-meet-prosody (= 1.0.2782-1)

Package: jibri
Version: 1.4.50-1
```

I'm more than happy to open a PR to make changes on README.md.

premanandmanimaran:
Hello 

I am trying to Install Jitsi Meet in AWS EC2 Instance to achieve the following
1. Desktop Sharing 
2. Audio / Video Call 
3. SIP Call
4. Call Recording 
5. Text Chat

Can you help me ? 


Anand

fanvictor:
hey. thanks for sharing.  How can I DM you?

Thanks.

[...]

PR #107, opened 2018-05-14
Installation disambiguation for AWS EC2 users

realnumber:
https://github.com/jitsi/jibri/issues/104


bbaldino:
@aaronkvanmeerten this look ok?

README.md
```diff
@@ -84,7 +87,7 @@ sudo apt-get install jibri
 
 
 ### User, group
-* Jibri is currently meant to be run as a regular system user. This example creatively uses username 'jibri' and group name 'jibri', but any user will do. This has not been tested with the root user.
+* Jibri is currently meant to be run as a regular system user. This example creatively uses username 'jibri' and group name 'jibri', but any user will do. This has been tested also with the root user.
```
aaronkvanmeerten: I would NOT recommend running jibri (or any daemon) as the root user.  We should explicitly suggest they do NOT do this, as we're not doing anything to ensure that jibri as root would be secure.

PR #113, opened 2018-05-21
Selenium rework

bbaldino:
Clean up some awkward bits in JibriSelenium

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/CallUrlInfo.kt
+++ b/src/main/kotlin/org/jitsi/jibri/CallUrlInfo.kt
@@ -17,18 +17,30 @@
 
 package org.jitsi.jibri
 
+import com.fasterxml.jackson.annotation.JsonIgnore
 import java.util.Objects
 
 /**
  * We assume the 'baseUrl' represents a sort of landing page (on the same
  * domain) where we can set the necessary local storage values.  The call
- * url will be created by joining [baseUrl] and [callName] with a "/"
+ * url will be created by joining [baseUrl] and [callName] with a "/".  If
+ * set, a list of [urlParams] will be concatenated after the call name with
+ * a "#" in between.
  */
 data class CallUrlInfo(
     val baseUrl: String = "",
-    val callName: String = ""
+    val callName: String = "",
+    private val urlParams: List<String> = listOf()
 ) {
-    val callUrl = "$baseUrl/$callName"
+    @get:JsonIgnore
+    val callUrl: String
+        get() {
+            return if (urlParams.isNotEmpty()) {
+                "$baseUrl/$callName#${urlParams.joinToString("&")}"
+            } else {
+                "$baseUrl/$callName"
+            }
+        }
 
     override fun equals(other: Any?): Boolean {
         return when {
@@ -40,6 +52,7 @@ data class CallUrlInfo(
     }
 
     override fun hashCode(): Int {
+        // Purposefully ignore urlParams here
         return Objects.hash(baseUrl.toLowerCase(), callName.toLowerCase())
     }
 }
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/jitsi/jibri/CallUrlInfo.kt
```diff
@@ -17,18 +17,30 @@
...
-    val callName: String = ""
+    val callName: String = "",
+    private val urlParams: List<String> = listOf()
 ) {
-    val callUrl = "$baseUrl/$callName"
+    @get:JsonIgnore
+    val callUrl: String
+        get() {
+            return if (urlParams.isNotEmpty()) {
+                "$baseUrl/$callName#${urlParams.joinToString("&")}"
```
virtuacoplenny: What happens if you call this even when urlParams is empty?

src/test/kotlin/org/jitsi/jibri/api/http/HttpApiTest.kt
```diff
@@ -66,6 +66,8 @@ class HttpApiTest : ShouldSpec() {
             override fun configure(): Application {
                 return ResourceConfig(object : ResourceConfig() {
                     init {
+                        // Uncommenting the following line can help with debugging any errors
+                        //property(LoggingFeature.LOGGING_FEATURE_LOGGER_LEVEL_SERVER, "WARNING")
```
virtuacoplenny: Do you think it's useful to leave this kind of debugging/test-help in the test itself. Does this tip not apply to tests in general?

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -236,6 +227,7 @@ class JibriSelenium(
             }
         }
         CallPage(chromeDriver).leave()
+        currCallUrl = null
```
virtuacoplenny: Is currCallUrl being stored for logging purposes?

src/main/kotlin/org/jitsi/jibri/CallUrlInfo.kt
bbaldino: it'll take the else below

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
bbaldino: yup, only for that which is annoying, but i think it's valuable to have in the logs.

src/test/kotlin/org/jitsi/jibri/api/http/HttpApiTest.kt
bbaldino: i thought so because i had an error here and it took some digging to find the right function to enable the debug logging for jersey (which was necessary to figure out what was going wrong)

src/main/kotlin/org/jitsi/jibri/CallUrlInfo.kt
virtuacoplenny: Sorry, I meant what if this didn't do a check for empty and just tried to append the urlParams no matter what?

src/test/kotlin/org/jitsi/jibri/api/http/HttpApiTest.kt
virtuacoplenny: And HttpApiTest is the only test that exercises jersey? And jersey cannot be abstracted away, I assume.

bbaldino: right, HttpApiTest actually uses this test jersey framework that makes it really go through the requests (rather than just calling the HttpApi methods directly) which is really nice because a lot of stuff happens there.

PR #118, opened 2018-05-24
Store recordings in separate directories

bbaldino:
THIS REPRESENTS A CHANGE IN JIBRI BEHAVIOR.  If the finalize script is successful (exits with code 0), then the recording will be deleted.  If it fails (non-zero exit code), the recording files will be moved into a subdirectory 'failed' within the recordings directory.


src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -88,50 +86,52 @@ data class RecordingMetadata(
...
         jibriSelenium.addStatusHandler(this::publishStatus)
     }
 
     override fun start(): Boolean {
-        if (!createIfDoesNotExist(fileRecordingParams.recordingDirectory, logger)) {
+        if (!createIfDoesNotExist(sessionRecordingDirectory, logger)) {
             return false
         }
-        if (!Files.isWritable(fileRecordingParams.recordingDirectory)) {
+        if (!Files.isWritable(sessionRecordingDirectory)) {
```
bbaldino: @bgrozev so i guess the idea here would be to change this to some helper which verified all permissions.  technically we'd need to do this at multiple points (since it could change).  i'm starting to think though that, issues there will fail with a proper stack (and we do clean up correctly), maybe we should just let that be the 'handling' of those scenarios.

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -224,9 +234,34 @@ class FileRecordingJibriService(private val fileRecordingParams: FileRecordingPa
...
             }
         } catch (e: Exception) {
             logger.error("Failed to run finalize script: $e")
         }
     }
+
+    private fun handleFinalizeSuccess() {
+        logger.info("Finalize finished successfully, deleting files")
+        try {
+            Files.walk(sessionRecordingDirectory)
```
bgrozev: Perhaps add this to FileUtils?

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -224,9 +234,34 @@ class FileRecordingJibriService(private val fileRecordingParams: FileRecordingPa
...
+                .sorted(Comparator.reverseOrder())
+                .map(Path::toFile)
+                .forEach { f -> f.delete() }
+        } catch (e: Exception) {
+            logger.error("Error deleting session recording directory", e)
+        }
+    }
+
+    private fun handleFinalizeError() {
+        logger.error("Error during finalize script, moving recording to failure directory")
```
bgrozev: Does that result in the recorded files being moved to `recordingDirectory/failed/sessionId/sessionId`?

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
bbaldino: Fixed

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
bbaldino: Nope, it uses the second argument as the entire target, which I found confusing as well.  I just double-checked if it could be done that way with the `Files.move` but it doesn't work :/

PR #132, opened 2018-06-25
Jibri no longer deletes/moves files after finalize failure/success

bbaldino:
Jibri will still execute the finalize script, passing the directory in
which the most recent recording has been written, but it will no longer
be taking responsibility for the recorded files after that point
(specifically, it will not check the return code of finalize--only log
it--and will not delete successfully finalized recordings nor move
unsuccessfully finalized ones).

This is because the new recommendation will be to have a separate
process handle the finished/failed recordings so that we don't block the
jibri while it has to do that work.


src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -171,7 +171,7 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
     fun leave(): Boolean {
         val result = driver.executeScript("""
             try {
-                return APP.conference._room.connection.disconnect();
+                return APP.conference._room.leave();
```
bbaldino: this was i tweak i made to better exit the meeting and make sure we send the presence-unavailable when we were going to handle that on the web to make it look like jibri left sooner.  i don't think we're going to end up doing that, but this change is probably still something we should have.

PR #147, opened 2018-09-12
Match average bitrate to image size (readable one)

Ark74:
As described here https://community.jitsi.org/t/improve-jibri-recording-resolution/14633
It would be very appreciated to increase the avg bit rate of the recordings to match the image size (12080x720), to be able to read text on the recorded streaming.
Cheers!

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/executor/FfmpegExecutor.kt
+++ b/src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/executor/FfmpegExecutor.kt
@@ -38,7 +38,7 @@ const val FFMPEG_RESTART_ATTEMPTS = 1
 data class FfmpegExecutorParams(
     val resolution: String = "1280x720",
     val framerate: Int = 30,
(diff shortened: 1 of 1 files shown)


bbaldino:
Thanks @Ark74.  I think we'll probably have to do some validation on our deployments that these settings work out alright, so that may take some time (@aaronkvanmeerten any thoughts?).  Another option which I think would be good would be to allow these settings to be overridden in the config file instead, and then we don't have to have hard-coded, unchangeable values in the code.

Ark74:
Sure.

> I think would be good would be to allow these settings to be overridden in the config file instead

That would be great too, at the end I guess the user might know the best options for their needs.
Cheers!

Ark74:
Just a heads up, if you go pass 17 (i.e. 15) it start to continuously show the "The participant is having connectivity issues" note.

My guess is that then jibri demands more quality (thus bandwidth) from the streamer, and if the bandwidth is not meet then the alert happens.

My upload is a sad 2 Mbps so 720p might be out of range.

PR #157, opened 2018-10-19
Jibri status overhaul

bbaldino:


bbaldino:
depends on https://github.com/jitsi/jitsi/pull/539, will need a pom update

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -149,11 +156,7 @@ class XmppApi(
      * Function to update outgoing [presence] stanza with current jibri status.
      */
     private fun updatePresenceStanza(presence: Presence) {
-        val jibriStatus = JibriStatusPacketExt()
-        jibriStatus.status =
-                if (jibriManager.busy()) JibriStatusPacketExt.Status.BUSY
-                else JibriStatusPacketExt.Status.IDLE
-        presence.addExtension(jibriStatus)
+        presence.addExtension(jibriStatusManager.overallStatus.toJibriStatusExt())
```
bgrozev: do you want to use addExtension or setExtension here?

bbaldino: thanks, will use the new `setChildExtension`

bbaldino: oh, actually this is a `Presence` which doesn't have a set

PR #159, opened 2018-10-23
Handle error cases and mark jibri as unhealthy

bbaldino:


src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -170,6 +171,12 @@ class JibriSelenium(
                 }
             } catch (t: Throwable) {
                 logger.error("Error while checking for empty call state: $t")
+                if (t is TimeoutException) {
+                    // We've found that a timeout here often implies chrome has hung so consider this as an error
+                    // which will result in this Jibri being marked as unhealthy
+                    logger.error("Javascript timed out, assuming chrome has hung")
+                    publishStatus(JibriServiceStatus.ERROR)
```
bgrozev: Do we publish any details for the error anywhere?

bbaldino: there are pieces of 'details' infrastructure in jibri, but it's not fully implemented yet...it's on the list for future (if we find it's needed) work.  right now details, where reported, are just logged in the jibri logs

PR #177, opened 2019-01-02
Handle chrome crashes

bbaldino:
make sure everywhere we're running executeScript we're handling an exception being thrown

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
@@ -226,25 +226,30 @@ class JibriSelenium(
     fun joinCall(callUrlInfo: CallUrlInfo, xmppCredentials: XmppCredentials? = null) {
         // These are all blocking calls, so offload the work to another thread
         TaskPools.ioPool.submit {
-            HomePage(chromeDriver).visit(callUrlInfo.baseUrl)
+            try {
+                HomePage(chromeDriver).visit(callUrlInfo.baseUrl)
 
-            val localStorageValues = mutableMapOf(
-                    "displayname" to jibriSeleniumOptions.displayName,
-                    "email" to jibriSeleniumOptions.email,
-                    "callStatsUserName" to "jibri"
-            )
-            xmppCredentials?.let {
-                localStorageValues["xmpp_username_override"] = "${xmppCredentials.username}@${xmppCredentials.domain}"
-                localStorageValues["xmpp_password_override"] = xmppCredentials.password
-            }
-            setLocalStorageValues(localStorageValues)
-            if (!CallPage(chromeDriver).visit(callUrlInfo.callUrl)) {
+                val localStorageValues = mutableMapOf(
+                        "displayname" to jibriSeleniumOptions.displayName,
+                        "email" to jibriSeleniumOptions.email,
+                        "callStatsUserName" to "jibri"
+                )
+                xmppCredentials?.let {
+                    localStorageValues["xmpp_username_override"] = "${xmppCredentials.username}@${xmppCredentials.domain}"
+                    localStorageValues["xmpp_password_override"] = xmppCredentials.password
+                }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -226,25 +226,30 @@ class JibriSelenium(
...
+                if (!CallPage(chromeDriver).visit(callUrlInfo.callUrl)) {
+                    stateMachine.transition(SeleniumEvent.FailedToJoinCall)
+                } else {
+                    startRecurringCallStatusChecks()
+                    addParticipantTracker()
+                    currCallUrl = callUrlInfo.callUrl
+                    stateMachine.transition(SeleniumEvent.CallJoined)
+                }
+            } catch (t: Throwable) {
+                logger.error("An error occurred while joining the call", t)
```
virtuacoplenny: Did you really want to wrap all of the above instead of targeting the chromedriver-specific lines?

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -267,7 +272,11 @@ class JibriSelenium(
...
         logger.info("Leaving web call")
-        CallPage(chromeDriver).leave()
+        try {
+            CallPage(chromeDriver).leave()
+        } catch (t: Throwable) {
+            logger.error("Error trying to leave the call", t)
+        }
         currCallUrl = null
         logger.info("Quitting chrome driver")
         chromeDriver.quit()
```
virtuacoplenny: I heard something about hung chromedrivers. Is there any force quitting logic necessary?

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -142,16 +142,27 @@ class FileRecordingJibriService(
...
     override fun stop() {
         logger.info("Stopping capturer")
         capturer.stop()
         logger.info("Quitting selenium")
-        val participants = jibriSelenium.getParticipants()
+        val participants = try {
+            jibriSelenium.getParticipants()
+        } catch (t: Throwable) {
+            logger.error("An error occurred while trying to get the participants list, proceeding with " +
+                    "and empty participants list", t)
```
virtuacoplenny: *an empty. Is it okay to continue with an empty list? Does that trigger the logic for auto-exiting?

src/main/kotlin/org/jitsi/jibri/service/impl/StreamingJibriService.kt
```diff
@@ -93,14 +94,19 @@ class StreamingJibriService(
...
+            jibriSelenium.addToPresence("mode", JibriIq.RecordingMode.STREAM.toString())
+            streamingParams.youTubeBroadcastId?.let {
+                if (!jibriSelenium.addToPresence("live-stream-view-url", "http://youtu.be/$it")) {
+                    logger.error("Error adding live stream url to presence")
+                }
             }
+            jibriSelenium.sendPresence()
+        } catch (t: Throwable) {
+            logger.error("Error while setting fields in presence", t)
+            publishStatus(ComponentState.Error(ErrorScope.SESSION, "Unable to set presence values"))
```
virtuacoplenny: Why did you prefer having start implement its own error handling versus having the caller of start implement it? Is it for the publishStatus behavior being the source of service status truth?

bbaldino: yeah, pretty much.  starts are async now, so all status is pushed via publishStatus

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
bbaldino: this is what we get from the API.

PR #196, opened 2019-05-09
Adds support for java 11.

damencho:


JonathanLennox:
I think the dependency changes should be covered by https://github.com/jitsi/jibri/pull/251, since it's Jersey that needs them?

What needed the maven-compiler-plugin change?

damencho:
> What needed the maven-compiler-plugin change?

I need to find it, but this was recommended for java11 support.

damencho:
Like this one: https://winterbe.com/posts/2018/08/29/migrate-maven-projects-to-java-11-jigsaw/#configure-plugins-for-java-11

JonathanLennox:
Looks like that advice is for _building_ with java11, not _running_ with it; that said, see https://github.com/jitsi/jibri/pull/252 .

damencho:
Thanks.

PR #200, opened 2019-06-11
Multi shard fix

bbaldino:
Think this is looking good, but we're gonna let it sit on 8x8 staging for a bit more to validate.


bbaldino:
Sorry, this also had a fix for adding a better exception that I forgot to open a separate PR for.  It's pretty isolated, but if it's confusing I can move it to its own PR (this is the first commit "add a bettter exception for when extracting call info from jid fails").

bbaldino:
cc @aaronkvanmeerten

Issue #162, opened 2018-11-04
Allow overriding/adding ffmpeg args.


bbaldino:
As mentioned [here](https://github.com/jitsi/jibri/issues/160#issuecomment-435178502), it could be useful to override some/pass additional ffmpeg args.  Getting into a situation where we have to 'merge' args would be a bit tricky, so maybe we want to allow them to be overridden entirely, or only allow overwriting certain arguments.

netaskd:
+1

PR #204, opened 2019-06-24
ffmpeg: change directly reference to an audio interface.

netaskd:
Now it just a link to .asound plug, that can be configurable without re-compilation kt files. 
This PR fixes issue described in https://github.com/jitsi/jibri/issues/162#issuecomment-460211648
We don't need change anything else for default behavior, because default interface already described here https://github.com/jitsi/jibri/blob/78258b63aeb24be9abf8bda518dcf328f1e98854/resources/debian-package/etc/jitsi/jibri/asoundrc#L32 We just link ffmpeg into this interface instead hardcoding inside kt file.


netaskd:
@jitsi-jenkins ICLA already signed.

bbaldino:
@aaronkvanmeerten any thoughts on this?

bbaldino:
@aaronkvanmeerten would we need deployment/infra changes to work with this change? or is it a drop-in replacement?

aaronkvanmeerten:
> @aaronkvanmeerten would we need deployment/infra changes to work with this change? or is it a drop-in replacement?

If it works as expected, it should already be functional.  This relies on the existing .asoundrc already having an alias called 'cloop' which can already be referenced.

bbaldino:
Tested this and audio looks good.

PR #207, opened 2019-07-03
take remote participant mute state into account when checking received

bbaldino:
media

turns out when people often _do_ have a recording running when everyone is audio and video muted (e.g. at the beginning of a call waiting for everyone to join).  this change takes into account the remote participants' mute state before error-ing out based on not receiving any media.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
@@ -172,7 +172,7 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
             is Number -> result.toInt()
             else -> {
                 logger.error("error running numRemoteParticipantsMuted script: $result ${result::class.java}")
-                1
+                0
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -154,6 +154,29 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
...
+                    .length;
+            } catch (e) {
+                return e.message;
+            }
+        """.trimMargin())
+        return when (result) {
+            is Number -> result.toInt()
+            else -> {
+                logger.error("error running numRemoteParticipantsMuted script: $result ${result::class.java}")
+                1
```
virtuacoplenny: Does the return of `numRemoteParticipantsMuted` equal `1` on error?

bbaldino: this value was leftover from another implementation of this method (where even 1 didn't feel great but couldn't think of a better default value).  now that we know this is numRemoteParticipants, i think 0 would be better.  will change that.

Issue #208, opened 2019-08-02
Chrome 76 shows in recorded file "chrome is being controlled by automated software"


netaskd:
## Description
---
Since chrome 76 the option "--disable-infobars" has been removed, which could previously be used to hide pop-up warnings in the Chrome interface (the CommandLineFlagSecurityWarningsEnabled rule has been proposed for hiding security-related warnings);
Option is here: https://github.com/jitsi/jibri/blob/845cc3274eeea14ad026417f5bfb2d44f06dcedb/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt#L134

## Possible Solution
---
the CommandLineFlagSecurityWarningsEnabled rule has been proposed for hiding security-related warnings. 
https://www.chromium.org/administrators/policy-list-3#CommandLineFlagSecurityWarningsEnabled

## Steps to reproduce
---
1. Update Chrome up to 76
2. Do record a conference
3. Open the recorded file
4. In the top of the recorded file you will see "chrome is being controlled by automated software"

# Environment details
---
Linux 3.10.0-957.27.2.el7.x86_64
ChromeDriver 75.0.3770.140
Google Chrome 76.0.3809.87
jibri 7.7.76-1

saghul:
Maybe this is also a solution? https://github.com/GoogleChrome/puppeteer/issues/2070#issuecomment-516694574

bbaldino:
Cool I'll try the solution from Puppeteer at some point.  @netaskd feel free to open a PR, we already have code passing  experimental options.

dbkr:
This just bit us when upgrading chrome. For the benefit of anyone who needs to work around this until jibri is updated, the flag @netaskd mentions can be used by creating the file `/etc/opt/chrome/policies/managed/managed_policies.json` with contents:
```
{
  "CommandLineFlagSecurityWarningsEnabled": false
}
```

netaskd:
Thanks @dbkr!
I think it's the best way to disable warnings, because:
1. setting via setExperimentalOption is unstable and can be changed in any next releases.
2. do not need re-compilation of the jibri package, can be changed on fly.

netaskd:
@bbaldino added PR #210

sergiistupkin:
Since "--disable-infobars" has been removed from Chrome 76, there is the solution: https://support.google.com/chrome/a/answer/187202 
This is works for me.

jbrundige:
Hello!

[...]

PR #210, opened 2019-08-07
The PR around issue #208

netaskd:
* remove option --disable-infobars as depricated since Chrome 76
* add information about creating chrome managed policy file to README

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
@@ -131,6 +131,7 @@ class JibriSelenium(
                 "--start-maximized",
                 "--kiosk",
                 "--enabled",
+                "--disable-infobars",
                 "--alsa-output-device=plug:amix",
                 "--autoplay-policy=no-user-gesture-required"
         )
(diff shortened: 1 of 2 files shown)


netaskd:
@jitsi-jenkins ICLA already signed.

bbaldino:
i just want to check with @aaronkvanmeerten on this.  any thoughts?

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -131,7 +131,6 @@ class JibriSelenium(
                 "--start-maximized",
                 "--kiosk",
                 "--enabled",
-                "--disable-infobars",
```
aaronkvanmeerten: Is this required to remove?  If we are trying to support both older and newer Chrome versions with this jibri build, we may want to continue to support the old command line parameters at least for some time.

bbaldino: yeah i'm tempted to say this should stay in as well for that reason.  it doesn't break anything in chrome 76, does it @netaskd ?

netaskd: you are right. It can be stayed without any problems in chrome 76

PR #211, opened 2019-08-07
Limit the amount of time we'll stay in the call if all participants are muted

bbaldino:


bbaldino:
Not gonna merge this yet, looking at maybe tweaking the implementation.

PR #212, opened 2019-08-21
Uses new versioning and tagging based on pom version and git describe.

damencho:


resources/jenkins/release.sh
```diff
@@ -3,43 +3,37 @@ set -e
...
-    MAJOR_VERSION=$((MAJOR_VERSION+1))
-    MINOR_VERSION=0
+if ! git rev-parse $TAG_NAME >/dev/null 2>&1
+then
+    git tag -a $TAG_NAME -m "Tagged automatically by Jenkins"
+	  git push origin $TAG_NAME
 else
-    echo "Error, unrecognized release type $MAJOR_MINOR.  Options are 'Major' and 'Minor'"
-    exit 1
+	echo "Tag: $TAG_NAME already exists."
```
saghul: shouldn't this abort the build and exit with error?

resources/jenkins/release.sh
```diff
@@ -3,43 +3,37 @@ set -e
...
+echo "Package version: ${VERSION}"
 
-git tag -a $NEW_VERSION -m "New $MAJOR_MINOR release"
+REV=$(git log --pretty=format:'%h' -n 1)
+
+# bulding the debian package expects the file target/jibri.jar
+mv target/jibri-${MVNVER}-jar-with-dependencies.jar target/jibri.jar
 
-dch -v "$NEW_VERSION.$BUILD_NUMBER-1" "Built from git. $REV"
+dch -v "${VERSION}-1" "Built from git. $REV"
```
saghul: shouldn't this have the short hash instead of the 1?

resources/jenkins/release.sh
damencho: Nope, so now there is no tag 8.0. The first time it sees it creates it and this will be ver 8.0-0. This means major.minor 8.0 and 0 commits after major.minor change. Let's say we rebase a PR with 3 commits and next version will be 8.0-3, 3 commits after major.minor change.
Once we want to bump minor, let's say, we do that in pom.xml and the script will create 8.1 tag and describe git command will continue counting commits for us and the deb package version will be 8.1-0 for the first 8.1 build and so on and so on.

resources/jenkins/release.sh
damencho: "-1" is the debian version. Normally that is used for packages in the official repos, when you just change something in the debian subfolder and you bump it, and normally that is the only source that is kept on debian version control servers, away from upstream. Updating upstream version always starts from "-1" debian version.
The actual hash from git describe command is only used when versioning the maven artifacts. There is a command to do the other way around I think ... hum... need to check how we can convert from 8.0-3 to hash.

damencho:
Please don't merge this yet! :) we still need to discuss some versioning stuff...

resources/jenkins/release.sh
```diff
@@ -3,43 +3,37 @@ set -e
...
+cd $WORKSPACE
 
-CURR_VERSION=`echo $DESCRIBE | awk '{split($0,a,"-"); print a[1]}'`
-echo "Got current git version $CURR_VERSION"
-MAJOR_VERSION=`echo $CURR_VERSION | awk '{split($0,a,"."); print a[1]}'`
-MINOR_VERSION=`echo $CURR_VERSION | awk '{split($0,a,"."); print a[2]}'`
+# Let's get version from maven
+MVNVER=`xmllint --xpath "/*[local-name()='project']/*[local-name()='version']/text()" pom.xml`
+TAG_NAME="v${MVNVER/-SNAPSHOT/}"
+echo $TAG_NAME
```
bbaldino: can we add some context to this print? like `echo "Current tag name: $TAG_NAME"`?

PR #221, opened 2019-09-19
Changes reload/start/restart.

damencho:
When reloading we want to restart the process when idle, so we exit with
 status 0 and systemd will restart the process.
When stopping graceful or not we want to exit the process with status as
 configured in systemd unit-file in RestartPreventExitStatus=255.


src/main/kotlin/org/jitsi/jibri/Main.kt
```diff
@@ -125,14 +125,24 @@ fun main(args: Array<String>) {
             exitProcess(0)
         }
     }
+    val gracefulShutdownHandler = {
+        logger.info("Jibri has been told to graceful shutdown, waiting to be idle before exiting")
+        jibri.executeWhenIdle {
+            logger.info("Jibri is idle and there are config file changes, exiting")
```
bbaldino: `"Jibri is idle and has been told to gracefully shutdown, exiting")`

src/main/kotlin/org/jitsi/jibri/Main.kt
```diff
@@ -125,14 +125,24 @@ fun main(args: Array<String>) {
...
+            // Exit with code 255 to indicate we do not want process restart
+            logger.info("Service stopped")
```
bbaldino: don't think this log is needed?

src/test/kotlin/org/jitsi/jibri/api/http/internal/InternalHttpApiTest.kt
```diff
@@ -46,20 +46,33 @@ class InternalHttpApiTest : ShouldSpec() {
...
+        "notifyConfigChanged" {
+            should("return a 200 and not invoke the config changed handler directly") {
+                var configChangedHandlerCalled = false
+                val configChangedHandler = {
+                    configChangedHandlerCalled = true
+                }
+                val internalHttpApi = InternalHttpApi({}, configChangedHandler, {})
+                val response = internalHttpApi.reloadConfig()
+                response.status shouldBe 200
+                configChangedHandlerCalled shouldBe false
```
bbaldino: realizing now that these tests should really verify that the handler does _eventually_ get called...but i'll go in and fix that later

src/main/kotlin/org/jitsi/jibri/Main.kt
damencho: ?

damencho: Got it

src/main/kotlin/org/jitsi/jibri/Main.kt
damencho: removed

bbaldino:
lgtm, but looks like you'll need to rebase to resolve some conflicts

damencho:
Rebased, should be fine now.

resources/debian-package/opt/jitsi/jibri/reload.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+curl -X POST http://127.0.0.1:3333/jibri/api/internal/v1.0/notifyConfigChanged
```
saghul: Can't we use SIGHUP instead? It's a more common approach to reloading the config of a running process.

damencho: I think there is no general mechanism in java to listen for those ...

saghul: ` For the signals SIGINT, SIGTERM, and SIGHUP the Java Runtime is terminated by invoking the equivalent of System.exit with a value of 128 + the signal number. Other signals are ignored. `

Ouch. Ignore what I said...

PR #223, opened 2019-09-23
Support a 'single use' mode in Jibri

bbaldino:
In this mode, a Jibri denotes its status as `EXPIRED` after a single use (note that a 'use' includes even an unsuccessful session).  This status is reflected in the health check but not the MUC (from Jicofo's view it will go `BUSY` and then not change).


bbaldino:
@aaronkvanmeerten the `EXPIRED` state change in health check is what the scripts will need to handle.

bbaldino:
ping @aaronkvanmeerten, want to make sure you see the health changes here, as it'll require changes in the health check script to handle this case.

bbaldino:
ping @aaronkvanmeerten, want to make sure you see the health changes here, as it'll require changes in the health check script to handle this case.

PR #225, opened 2019-09-30
Adds auto-answer mode to sipgw, activated through http-api.

damencho:


src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -222,6 +222,13 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
                 return e.message;
             }
         """.trimMargin())
+
+        // Let's wait till we are alone in the room
+        // (give time for the js Promise to finish before quiting selenium)
+        WebDriverWait(driver, 2).until {
```
bbaldino: is there something specific driving this change?

src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
```diff
@@ -66,7 +66,9 @@ class PjsuaClient(private val pjsuaClientParams: PjsuaClientParams) : SipClient(
                 "--id", "${pjsuaClientParams.sipClientParams.displayName} <sip:jibri@127.0.0.1>",
                 "--config-file", CONFIG_FILE_LOCATION,
                 "--log-file", "/tmp/pjsua.out",
-                "sip:${pjsuaClientParams.sipClientParams.sipAddress}"
+                "--max-calls=1",
+                if (pjsuaClientParams.sipClientParams.autoAnswer)
+                    "--auto-answer=200" else "sip:${pjsuaClientParams.sipClientParams.sipAddress}"
```
bbaldino: am i  reading this right that this  will put it  in a mode  where, instead of calling it out it will go into a mode where it waits for  an incoming call?  is the  `200` some kind of timeout?

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
damencho: Yes, when using sipgw you see the participant and leave does not work. It is not noticeable with the recorder as the participant is not visible.

src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
damencho: 200 is the sip code for the auto-answer:
```
--auto-answer=code  Automatically answer incoming calls with code (e.g. 200)
```
Yes, the idea is not to trigger this functionality from the xmpp side, but from the sip side for incoming sip calls.

bbaldino: so is there some kind  of timeout?  will jibri  just sit like this forever?

damencho: Good point I will add a timeout, thanks.

damencho:
I've added a timer which depends on https://github.com/jitsi/pjproject/pull/1
But I found few more cases which I will work on:
1. For some reason when pjsua timeouts and exits the selenium part does not exit the meeting as it is trying to kill pjsua.
2. We need to make jibri enter the room and be able to wait ... even 30 miniutes for someone to join (make it configurable this timeout), as currently, I think it does not do that ... but I can also be wrong.
3. We need in case of auto-answer make the http request return only when pjsua is started, which is after we join the selenium part of the call.

bbaldino:
> I've added a timer which depends on [jitsi/pjproject#1](https://github.com/jitsi/pjproject/pull/1)
> But I found few more cases which I will work on:
> 
> 1. For some reason when pjsua timeouts and exits the selenium part does not exit the meeting as it is trying to kill pjsua.

I may be able to help with this if you've got some logs of when it happens.

> 2. We need to make jibri enter the room and be able to wait ... even 30 miniutes for someone to join (make it configurable this timeout), as currently, I think it does not do that ... but I can also be wrong.

Does this mean Jibri needs to be able to join a room and wait there alone for this amount of time?  If so, then yes we'll need to add some flexibility to the empty call check in the case of SIP gw.

> 3. We need in case of auto-answer make the http request return only when pjsua is started, which is after we join the selenium part of the call.

Jibri needs to make an HTTP request? What request?

damencho:
> Does this mean Jibri needs to be able to join a room and wait there alone for this amount of time? If so, then yes we'll need to add some flexibility to the empty call check in the case of SIP gw.

Yep, that is correct. We were doing same and for jigasi, basically, all dial-ins need to be able to be waiting for others to join.

> Jibri needs to make an HTTP request? What is the request?
No, it is not jibri. The part that will be starting this logic from this PR needs to wait for pjsua to start listening so it can forward the sip call there...

bbaldino:
> > Does this mean Jibri needs to be able to join a room and wait there alone for this amount of time? If so, then yes we'll need to add some flexibility to the empty call check in the case of SIP gw.
> 
> Yep, that is correct. We were doing same and for jigasi, basically, all dial-ins need to be able to be waiting for others to join.
> 

Ok, I have an idea for how we can add some customization for the timeout there.  I will try something.

> > Jibri needs to make an HTTP request? What is the request?
> > No, it is not jibri. The part that will be starting this logic from this PR needs to wait for pjsua to start listening so it can forward the sip call there...

damencho:
> > 1. For some reason when pjsua timeouts and exits the selenium part does not exit the meeting as it is trying to kill pjsua.
> 
> I may be able to help with this if you've got some logs of when it happens.

There are no logs, for some reason the exception is swallowed. The problem is that I'm running it on macosx and `Runtime.getRuntime().exec("kill -s SIGINT $pid")` throws an exception that there is no such command.

I've added a try/catch/finally around [`pjsuaClient.stop()`](https://github.com/jitsi/jibri/blob/master/src/main/kotlin/org/jitsi/jibri/service/impl/SipGatewayJibriService.kt#L95), to make sure selenium is always stopped.

bbaldino:
> > > 1. For some reason when pjsua timeouts and exits the selenium part does not exit the meeting as it is trying to kill pjsua.
> > 
> > 
> > I may be able to help with this if you've got some logs of when it happens.
> 
> There are no logs, for some reason the exception is swallowed. The problem is that I'm running it on macosx and `Runtime.getRuntime().exec("kill -s SIGINT $pid")` throws an exception that there is no such command.
> 
> I've added a try/catch/finally around [`pjsuaClient.stop()`](https://github.com/jitsi/jibri/blob/master/src/main/kotlin/org/jitsi/jibri/service/impl/SipGatewayJibriService.kt#L95), to make sure selenium is always stopped.

It's weird you see this, as we use the same method for stopping ffmpeg and i've run that a lot on my mac and never seen this kill issue.  We do swallow logs in places because we're running the task in a threadpool so need to capture it there (it's async, so the exception wouldn't bubble up to any synchronous code anyway), so I'd expect to at least see the log there. That being said, we should be handling an exception here, I just hadn't seen it before so we don't have logic for it.

damencho:
> There are no logs, for some reason the exception is swallowed. The problem is that I'm running it on macosx and `Runtime.getRuntime().exec("kill -s SIGINT $pid")` throws an exception that there is no such command.

This was to a local change in the environment vars when running the project :(

damencho:
I think everything I found while testing is now handled. If you can give a final look.

src/main/kotlin/org/jitsi/jibri/service/impl/SipGatewayJibriService.kt
```diff
@@ -85,14 +88,24 @@ class SipGatewayJibriService(
     override fun start() {
         jibriSelenium.joinCall(
             sipGatewayServiceParams.callParams.callUrlInfo.copy(urlParams = SIP_GW_URL_OPTIONS))
-        whenever(jibriSelenium).transitionsTo(ComponentState.Running) {
+
+        // when in auto-answer mode we want to start as quick as possible as
+        // we will be waiting for a sip call to come
+        if (sipGatewayServiceParams.sipClientParams.autoAnswer) {
             pjsuaClient.start()
+        } else {
```
bbaldino: so it's ok here if the pjsua client starts (and maybe connects?) before the browser has started/joined the call? won't we potentially forward the desktop background or something there?

src/main/kotlin/org/jitsi/jibri/service/impl/SipGatewayJibriService.kt
```diff
@@ -85,14 +88,24 @@ class SipGatewayJibriService(
...
+                pjsuaClient.start()
+            }
         }
     }
 
     override fun stop() {
         processMonitorTask?.cancel(false)
-        pjsuaClient.stop()
-        jibriSelenium.leaveCallAndQuitBrowser()
+        try {
```
bbaldino: this try/catch should no longer be needed, right?

src/main/kotlin/org/jitsi/jibri/service/impl/SipGatewayJibriService.kt
damencho: We will see, in the past, we were setting some branded background image so it doesn't look weird.

The point is that loading the web can take a while and we will be trying to send a sip call to be established and we expect port 5060 to be open.

src/main/kotlin/org/jitsi/jibri/service/impl/SipGatewayJibriService.kt
damencho: Well, I think for safety it should be there. 
I experienced one issue where the chrome stays opened when kill fails for some reason. 
We know how bad this can be from the recent experience, so I was thinking we better keep the try-finally code :)

bbaldino: if that call is throwing, i'd rather we make sure it's handled appropriately there (along with the other exceptions) so that if it's called from elsewhere we don't run into the same problem.

bbaldino: if we do want an exception catch-all, we should put it in `JibriSubprocess`

damencho: But there is no catch-all here. It is just try-finally, just to be sure that whatever happens quitting selenium will be called.

bbaldino: sorry, what i mean is that the call to stop shouldn't throw, as it isn't the error-propagation model being used here.  it was a mistake on my part not to be more thorough when checking what exceptions were possible in `JibriSubprocess#stop`, but here we are.  i don't want to put a finally here, as it suggests this code should potentially handle an exception when it shouldn't.  let's remove this part and i will take another pass at the calls in `JibriSubprocess#stop` to make sure all exceptions are handled correctly.

damencho: Ok, I dropped that commit.

bbaldino: thanks, merged

PR #226, opened 2019-10-04
Custom empty call timeout

bbaldino:
this will allow a custom call-empty-timeout value for Jibri, so for SIP gw cases we can increase it.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
@@ -158,7 +158,7 @@ class JibriSelenium(
         // for media being received, since the call being empty will also mean Jibri is not receiving media but should
         // not cause Jibri to go unhealthy (like not receiving media when there are others in the call will).
         callStatusChecks = listOf(
-            EmptyCallStatusCheck(),
+            EmptyCallStatusCheck(logger),
             MediaReceivedStatusCheck(logger)
         )
         stateMachine.onStateTransition(this::onSeleniumStateChange)
--- a/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
@@ -78,7 +78,11 @@ data class JibriSeleniumOptions(
      * Chrome command line flags to add (in addition to the common
      * ones)
      */
-    val extraChromeCommandLineFlags: List<String> = listOf()
+    val extraChromeCommandLineFlags: List<String> = listOf(),
+    /**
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -158,7 +160,7 @@ class JibriSelenium(
         // for media being received, since the call being empty will also mean Jibri is not receiving media but should
         // not cause Jibri to go unhealthy (like not receiving media when there are others in the call will).
         callStatusChecks = listOf(
-            EmptyCallStatusCheck(),
+            EmptyCallStatusCheck(logger, callEmptyTimeout = jibriSeleniumOptions.emptyCallTimeout),
```
damencho: I'm not sure, how we will set this Duration in the config as seconds/milliseconds, shouldn't that be defined and described with some doc.

bbaldino: oh, i didn't know you wanted to set this from a config file, i thought you just wanted the sip gateway service to pass a different timeout value?

damencho: Hum, it will not be config but json that comes from http api rest call

bbaldino: what actually comes in that rest call, it's how long jibri should wait in an empty call before leaving?  how does it know how long it should wait?

damencho: The rest call is 
```
{
    "sessionId": 123456789,
    "callParams": {
		"callUrlInfo": {
			"baseUrl": "https://meet.jit.si",
			"callName": "sipgwtest"
		}
	},
    "sinkType": "gateway",
    "sipClientParams": {
        "displayName": "+11111111",
        "autoAnswer": true
    }
}
```
I was thinking we can add `emptyCallTimeout : 1800000` in this json to indicate that after the incoming sip call is connected we can wait 30 mins in the conference alone before giving up ...

bbaldino: do we need to have this in the rest call, though?  i mean, does it make sense to have that code determine how long we wait as opposed to just having some time set in jibri?  would this time vary from call to call? what would make it shorter or longer?

damencho: Yep, there is no point of being there. But I think being in the main.json sounds right to me...

damencho: I will work on those things today ...

PR #228, opened 2019-10-17
Subprocess stop fixes

bbaldino:
this adds some better handling for a couple exception scenarios around stopping a subprocess

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/util/JibriSubprocess.kt
+++ b/src/main/kotlin/org/jitsi/jibri/util/JibriSubprocess.kt
@@ -17,6 +17,7 @@
 package org.jitsi.jibri.util
 
 import org.jitsi.jibri.util.extensions.error
+import java.time.Duration
 import java.util.concurrent.Future
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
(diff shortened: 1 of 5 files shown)


bbaldino:
Want to double check the pid change on an actual jibri machine, but it worked fine on my mac.

bbaldino:
tested the pid change on a jibri machine, looks good.

src/test/kotlin/org/jitsi/jibri/util/ProcessWrapperTest.kt
```diff
@@ -161,5 +163,29 @@ internal class ProcessWrapperTest : ShouldSpec() {
...
+                }
+
+            }
+            "when the runtime throws RuntimeException" {
+                whenever(runtime.exec(anyString())).thenAnswer { throw RuntimeException() }
+                should("handle it correctly") {
+                    processWrapper.stopAndWaitFor(Duration.ofSeconds(10)) shouldBe false
+                }
+            }
+        }
```
virtuacoplenny: destroyForciblyAndWaitFor tests?

src/main/kotlin/org/jitsi/jibri/util/JibriSubprocess.kt
```diff
@@ -55,31 +56,32 @@ class JibriSubprocess(
...
+
+    fun stop() {
+        logger.info("Stopping $name process")
+        processStatePublisher?.stop()
+
+        process?.apply {
+            if (!stopAndWaitFor(Duration.ofSeconds(10))) {
+                logger.error("Error trying to gracefully stop $name, destroying forcibly")
+                if (!destroyForciblyAndWaitFor(Duration.ofSeconds(10))) {
+                    logger.error("Error trying to destroy $name forcibly")
```
virtuacoplenny: Follow up PR maybe to handle if forced stop failed.

src/test/kotlin/org/jitsi/jibri/util/ProcessWrapperTest.kt
bbaldino: fixed

src/main/kotlin/org/jitsi/jibri/util/JibriSubprocess.kt
bbaldino: yeah i'm thinking we should go to an error state here.  will do that in another PR

PR #233, opened 2019-11-28
Updates jibri pjsua part

damencho:


resources/debian-package/etc/jitsi/jibri/asoundrc
```diff
@@ -19,23 +19,34 @@ pcm.aduplex {
   capture.pcm "asnoop"
 }
 
-# ------------------------------------------------------
-# for jack alsa_in and alsa_out: looped-back signal at other ends
-pcm.ploop {
-  type plug
-  slave.pcm "hw:Loopback,1,1"
+# playback PCM device: using loopback subdevice 0,0
```
saghul: Let's drop the comments, I didn't update them bedore I gave you the file...

resources/debian-package/etc/jitsi/jibri/pjsua.config
```diff
@@ -0,0 +1,31 @@
...
+--dis-codec opus
+--add-codec pcmu
+--add-codec pcma
+--add-codec speex
+--add-codec G722
+--add-codec opus
+--no-vad
+--ec-tail 0
+--quality 10
+#--auto-answer-timer=30
```
saghul: why are these commened out?

resources/debian-package/etc/jitsi/jibri/pjsua.config
```diff
@@ -0,0 +1,31 @@
...
+#--auto-answer=200
+#--max-calls=1
+--no-stderr
+--log-file=/tmp/pjsua.out
```
saghul: /var/log/jibri/pjsua.log while we are at it?

src/main/kotlin/org/jitsi/jibri/sipgateway/SipClient.kt
```diff
@@ -34,7 +34,17 @@ data class SipClientParams(
      * Whether auto-answer is enabled, if it is, the client will listen for
      * incoming invites and will auto answer the first one.
      */
-    val autoAnswer: Boolean = false
+    val autoAnswer: Boolean = false,
```
saghul: Are we planning on supporting the non auto-answer case?

resources/debian-package/etc/jitsi/jibri/pjsua.config
damencho: If you use sip-jibri from the xmpp side, not sure how it will act ...

src/main/kotlin/org/jitsi/jibri/sipgateway/SipClient.kt
damencho: This is the xmpp side triggering this, as it creates a sip call.

resources/debian-package/etc/jitsi/jibri/asoundrc
```diff
@@ -1,41 +1,45 @@
...
   ipc_key 219345
   slave.pcm "hw:Loopback,0,0"
 }
 
-# capture PCM device: using loopback subdevice 0,1
 pcm.asnoop {
   type dsnoop
   ipc_key 219346
-  slave.pcm "hw:Loopback,0,1"
+  slave.pcm "hw:Loopback_1,1,0"
```
saghul: wait, why this change?

resources/debian-package/etc/jitsi/jibri/pjsua.config
saghul: But then it won't work on the other case... I guess it's ok for now, but maybe we should move those to back the Java side so it can be made functional both ways without changes?

`--max-calls` can stay

resources/debian-package/etc/systemd/system/jibri-xorg.service
```diff
@@ -6,7 +6,7 @@ After=network.target
 User=jibri
 Group=jibri
 Environment=DISPLAY=:0
-ExecStart=/usr/bin/Xorg -nocursor -noreset  +extension RANDR +extension RENDER -logfile /tmp/xorg.log  -config /etc/jitsi/jibri/xorg-video-dummy.conf :0
+ExecStart=/usr/bin/Xorg -nocursor -noreset  +extension RANDR +extension RENDER -logfile /var/log/xorg.log  -config /etc/jitsi/jibri/xorg-video-dummy.conf :0
```
saghul: Since this X server is for Jibri I'd put the log file in /var/log/jitsi/jibri

PR #238, opened 2020-01-07
Fixes audio device name for recording.

damencho:


bbaldino:
Could you include a description of this change?

PR #242, opened 2020-02-14
add logging, surround retrieving chromedriver logs with try/catch

bbaldino:


src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -272,17 +272,23 @@ class JibriSelenium(
     }
 
     fun leaveCallAndQuitBrowser() {
+        logger.info("Leaving call and quitting borwser")
```
aaronkvanmeerten: Nit: borwser typo

PR #253, opened 2020-03-24
limit the recording filename length

bbaldino:


src/main/kotlin/org/jitsi/jibri/sink/impl/FileSink.kt
```diff
@@ -18,20 +18,29 @@
...
 class FileSink(recordingsDirectory: Path, callName: String, extension: String = "mp4") : Sink {
     val file: Path
     init {
-        val currentTime = LocalDateTime.now()
-        val formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd-HH-mm-ss")
-        val filename = "${callName}_${currentTime.format(formatter)}.$extension"
+        val timestampStr = LocalDateTime.now().format(TIMESTAMP_FORMATTER)
+        // How much of the call name we'll use is determined by how many characters are leftover
+        // counting the timestamp and underscore separator
+        val maxCallNameLength = MAX_FILENAME_LENGTH - timestampStr.length - 1
```
bgrozev: Is this -1 to account for the "_"? I don't know whether it's important to us exactly 125 charactrs, or just any limit. I don't understand why the extension is considered special.

I would suggest counting the extension, and doing something like
```
val postfix = "_${currentTime.format(formatter)}.$extension"
val filename = "${callName.take(MAX_FILENAME_LENGTH - postfix.length)}${postfix}"
```

src/test/kotlin/org/jitsi/jibri/sink/impl/FileSinkTest.kt
```diff
@@ -41,5 +43,27 @@ internal class FileSinkTest : ShouldSpec() {
                 sink.format shouldBe "ext"
             }
         }
+        "when created with a really long call name" {
+            val reallyLongCallName = String.randomAlphas(200)
+            val sink = FileSink(fs.getPath("/tmp/xxx"), reallyLongCallName, "ext")
+            should("not generate a filename longer than the max file length") {
+                sink.file.baseName().length shouldBe 125
```
bgrozev: Use the `MAX_FILENAME_LENGTH` constant?

bbaldino: Oops, meant to do that. Will fix.

src/main/kotlin/org/jitsi/jibri/sink/impl/FileSink.kt
bbaldino: I tried many iterations of this entire thing.  The `-1` is indeed to cover the `_`.  125 itself is arbitrary.  I wanted this to at least roughly look like an implementation of enforcing a true filesize limit (rather than an arbitrary size) and I googled around and _think_ the extension doesn't count against that.  

But, I like your suggestion so will use that.

PR #270, opened 2020-04-16
Port 5222 has to be opened on the Jitst Meet server

goappycodes:
I believe it is very important to let users know that Port 5222 has to be opened on the Jitst Meet server to allow inbound connections. Might save a few days effort for people :)


bbaldino:
I'm not sure this feels like it belongs in the Jibri README, does nothing else require that port to be accessible?

kpeiruza:
@bbaldino everything connecting remotely to Prosody needs it, i.e. videobridges, jigasi or jicofo if it's not on the same machine.

fremzy:
> @bbaldino everything connecting remotely to Prosody needs it, i.e. videobridges, jigasi or jicofo if it's not on the same machine.

True, but if everything is hosted on the same server, nothing else appears to require opening that port. It appears Jibri needs that port open regardless of where it's installed - same server as JMS or separate standalone server.

PR #297, opened 2020-06-01
Xmpp port setting added to config

slymnshn:
Port property added to config


bbaldino:
Not opposed to this, but a few things:

1. It looks like you've got more changes in here than just the port (the changes in `JibriSelenium.kt`)
1. Believe this will have issues if no ported is specified: we'll want it to default to the default port

bbaldino:
Thanks, can you add a unit test?

bbaldino:
Closing this as it's based on the old (deprecated) config.

PR #302, opened 2020-06-19
Don't respond with 'pending' if Jibri is busy

bbaldino:
We ran into the following scenario:

* The same Jicofo sent start requests to the same Jibri for 2 different conferences (right after one another, before the Jibri had a chance to transition to 'BUSY' from the first request).
* Jibri responds to both requests immediately with a `Pending` status.
* Jibri services the first request, and sends an error (`Busy`) for the second request
* Since that error is a new IQ, Jicofo routes it based on the nickname of the Jibri it came from.  It does this by asking each conference which Jibri it is talking with.  Since there are _two_ conferences talking to this Jibri, it routes it to the first one it finds (which may be the wrong one).  This causes a mismatch in between what Jibri is doing and what Jicofo thinks it is doing.

This fix delays the initial `Pending` response by deferring when we launch the new service into a separate thread until after we've determined that the Jibri isn't already busy.

### DIFF
--- a/src/test/kotlin/org/jitsi/jibri/api/xmpp/XmppApiTest.kt
+++ b/src/test/kotlin/org/jitsi/jibri/api/xmpp/XmppApiTest.kt
@@ -20,7 +20,6 @@ package org.jitsi.jibri.api.xmpp
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.nhaarman.mockitokotlin2.any
 import com.nhaarman.mockitokotlin2.argumentCaptor
-import com.nhaarman.mockitokotlin2.doAnswer
 import com.nhaarman.mockitokotlin2.mock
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/jitsi/jibri/api/xmpp/XmppApiTest.kt
```diff
@@ -150,6 +151,19 @@ class XmppApiTest : ShouldSpec() {
                         }
                     }
                 }
+                "and jibri is busy" {
+                    whenever(jibriManager.startFileRecording(any(), any(), any(), any())).thenAnswer {
```
paweldomas, line 171: how does `whenever` `thenAnswer` work? does it mock the method to always throw?

bbaldino: Yeah

PR #304, opened 2020-06-26
allow custom options for chrome driver

### LINKED ISSUE #298: custom option for chromedriver
Hi, 
first thanks for providing such a cool jitsi companion!
I have a special use case where I need to enable a chrome extension on startup, my best guess was to add `--load-extensions=/path/to/extension/folder` flag in `chromeOptions` variable. Do you have any other/better idea?
I will provide a pull request soon, I hope it will make easier to understand my point ;)

stregouet:
This change will allow to customize chrome arguments from outside jibri code. This is done with an extra environment variable 
`CHROME_DRIVER_EXTRA_OPTS`.

closes #298

### DIFF
--- a/resources/debian-package/etc/jitsi/jibri/config.json
+++ b/resources/debian-package/etc/jitsi/jibri/config.json
@@ -6,6 +6,8 @@
     "recording_directory":"/tmp/recordings",
     // The path to the script which will be run on completed recordings
     "finalize_recording_script_path": "/path/to/finalize_recording.sh",
+    // extra options for chrome driver
+    "chrome_extra_opts": [],
     "xmpp_environments": [
         {
             // A friendly name for this environment which can be used
--- a/src/main/kotlin/org/jitsi/jibri/JibriManager.kt
+++ b/src/main/kotlin/org/jitsi/jibri/JibriManager.kt
@@ -21,6 +21,8 @@ import org.jitsi.jibri.config.JibriConfig
 import org.jitsi.jibri.config.XmppCredentials
 import org.jitsi.jibri.health.EnvironmentContext
 import org.jitsi.jibri.selenium.CallParams
+import org.jitsi.jibri.selenium.JibriSelenium
+import org.jitsi.jibri.selenium.JibriSeleniumOptions
 import org.jitsi.jibri.service.JibriService
 import org.jitsi.jibri.service.JibriServiceStatusHandler
 import org.jitsi.jibri.service.ServiceParams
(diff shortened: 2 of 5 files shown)


bbaldino:
I don't want to use an env variable for this (at least, not with the current config) as it's inconsistent with the rest of the config.  I'll be doing some updates in Jibri though, so I'll keep this use case in mind.

stregouet:
ok, so I can modify `JibriConfig.kt` to accept extra configuration instead of using environment variable, what do you think?

This will make a much bigger change, but it's definitely not impossible ;)

bbaldino:
Yeah, that would be the way forward I think.  Eventually I hope to move Jibri over to a config library we use elsewhere which supports overriding config values with environment variables, but even then we'll still want it in the config (to be able to be overridden).

bbaldino:
Thanks for the tweak...going to try and look at this this week.

bbaldino:
(Haven't forgotten, just haven't gotten to it yet)

bbaldino:
I've got some work in progress that's going to change config in Jibri, so want to hold off a bit on this.

stregouet:
ok, do you have any idea when your work will be available? 
tell me if I can help!

bbaldino:
> ok, do you have any idea when your work will be available?
> tell me if I can help!

I hope in the next week.  If you'd be up for testing on your env that would be helpful: Jibri will migrate to a new config file but with backwards compatibility support for the old one (for some time).  So testing would be helpful.

bbaldino:
> ok, do you have any idea when your work will be available?
> tell me if I can help!

I've now got a PR here: https://github.com/jitsi/jibri/pull/324.  If you could try it out and give feedback on the [migration guide](https://github.com/bbaldino/jibri/blob/new_config/doc/config_migration_guide.md) that would be helpful.

bbaldino:
I'm still waiting on a review to merge the config migration PR, but I created a branch on top of that to do the extra chrome options [here](https://github.com/bbaldino/jibri/tree/chrome_opts).  With that code, if you wanted to add another flag you'd put this in `application.conf`:
```hocon
jibri {
    chrome {
        flags = ${jibri.chrome.flags} [
            "--some-other-flag"
        ]
    }
}
```

If you wanted to overwrite the flags altogether, you'd do:
```hocon
jibri {
    chrome {
        flags = [
            "--new-flag-one",
            "--new-flag-two"
        ]
    }
}
```

bbaldino:
It's now merged: https://github.com/jitsi/jibri/pull/327

PR #305, opened 2020-06-30
Upgrade kotest (kotlintest)

bbaldino:


src/test/kotlin/org/jitsi/jibri/helpers/Helpers.kt
```diff
@@ -43,6 +43,10 @@ fun <T> within(duration: Duration, func: () -> T): T {
 }
 
 fun Int.seconds(): Duration = Duration.ofSeconds(this.toLong())
+val Int.seconds: Duration
```
bgrozev, line 46: Do we need both the `fun` and the `val`?

bbaldino: Nope, fixed.

PR #315, opened 2020-07-08
Add a Jibri status 'webhook'

bbaldino:
This PR adds a new 'webhook' for consumers to be able to get updates on Jibri's status.  There are accompanying configuration changes, but all new config keys have default values so upgrades shouldn't break (the `jibri_id` key will likely have its default value removed eventually).

This implementation is more of a 'pseudo' webhook, as there is no dynamic registration (webhook consumers must have their base URLs configured in the config file) and the callback URL is not customizable, instead it is assumed that, at the configured base URL, a 'contract' (currently of only a single method) defined by Jibri is implemented.  There's a README containing more information.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/webhooks/v1/README.md
+++ b/src/main/kotlin/org/jitsi/jibri/webhooks/v1/README.md
@@ -1,0 +1,46 @@
+# Jibri 'Webhooks'
+
+Jibri supports being configured with a list of base URLs on which it will hit certain endpoints with data.  These are
+not "true" webhooks as the endpoints are hard-coded, instead Jibri defines a "contract" which it expect a subscriber
+to implement at the given base URL.  Information about this contract is below.
+
+### Status updates
+Jibri pushes status updates consisting of its "busy status" (whether it is busy or idle) and its health.
+
+URL: `/v1/status`
+
+method: `POST`
+
+Data constraints:
+```$json
+{
+    "jibriId":"[String]",
+    "status":{
+        "busyStatus":"[a String value of ComponentBusyStatus: (BUSY|IDLE|EXPIRED)]",
+        "health": {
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/jitsi/jibri/webhooks/v1/README.md
```diff
@@ -0,0 +1,46 @@
+# Jibri 'Webhooks'
+
+Jibri supports being configured with a list of base URLs on which it will hit certain endpoints with data.  These are
+not "true" webhooks as the endpoints are hard-coded, instead Jibri defines a "contract" which it expect a subscriber
+to implement at the given base URL.  Information about this contract is below.
+
+### Status updates
+Jibri pushes status updates consisting of its "busy status" (whether it is busy or idle) and its health.
+
+URL: `/v1/status`
```
jmacelroy, line 10: Just this morning I thought that we need to somehow assert that an "real" Jibri is triggering these webhooks. We will probably want to use ASAP JWTs like we've done for other things. Shouldn't block this PR but reading this reminded me of it.

src/main/kotlin/org/jitsi/jibri/webhooks/v1/README.md
```diff
@@ -0,0 +1,46 @@
+# Jibri 'Webhooks'
+
+Jibri supports being configured with a list of base URLs on which it will hit certain endpoints with data.  These are
+not "true" webhooks as the endpoints are hard-coded, instead Jibri defines a "contract" which it expect a subscriber
+to implement at the given base URL.  Information about this contract is below.
+
+### Status updates
+Jibri pushes status updates consisting of its "busy status" (whether it is busy or idle) and its health.
```
jmacelroy, line 8: How often does it push these and is that configurable? I think the answer is every 60 seconds and on any state change but it helps to call it out in this readme.

PR #324, opened 2020-07-22
Migrate to new config

bbaldino:
This PR adds support in Jibri for a new mechanism for retrieving configuration.


bbaldino:
> New installs will still use old config?

No, I've got a PR in progress in infra to generate a new config file...when doing that earlier I saw that the launch script is pulled from this repo (in the debian dir) so will add another param to point to the new config there (debian package will still look at old config file and will prioritize any settings there before the new config)

damencho:
Can we make it that new installs directly provide the new config only? 
Old installs that had updated just the code but not the config use old if exist or new one ...

bbaldino:
> Can we make it that new installs directly provide the new config only?
> Old installs that had updated just the code but not the config use old if exist or new one ...

Yeah, that's the plan.  New installs will deploy a new config only, but my thought was we'd still include the `--config "/etc/jitsi/jibri/config.json"` argument [in launch.sh](https://github.com/jitsi/jibri/blob/master/resources/debian-package/opt/jitsi/jibri/launch.sh#L3) in case the user had an old config file.  Or do you think that's not necessary?  The goal is just to not break people's installs when they upgrade.

README.md
```diff
@@ -89,7 +89,9 @@ sudo usermod -aG adm,audio,video,plugdev jibri
 ```
 
 ### Config files
-* Edit the `config.json` file (installed to `/etc/jitsi/jibri/config.json` by default) appropriately.
+* Edit the `application.conf` file (installed to `/etc/jitsi/jibri/application.conf` by default) appropriately.  You can look at
```
bgrozev, line 92: Shoud we install to `/etc/jitsi/jibri/jibri.conf`?

doc/config_migration_guide.md
```diff
@@ -0,0 +1,169 @@
...
+            "call_login": {
+                "domain": <String>,
+                "username": <String>,
+                "password": <String>
+            },
+            "room_jid_domain_string_to_strip_from_start": <String>,
+            "usage_timeout": <number>
+        },
+        { <more> }
+    }
```
bgrozev, line 125: Missing `]}` (I think). It confuses the github interface, don't know if it affects te md file itself

src/main/kotlin/org/jitsi/jibri/Main.kt
```diff
@@ -113,13 +66,21 @@ fun main(args: Array<String>) {
             }
         }
     }
+    val jibriId = configSupplier<String> {
```
bgrozev, line 69: Intentionally not using the `config` delegate?

bbaldino: So I thought it was because you can't use delegates on non-members, but I just checked and `by lazy` does work, but the config ones complain.  I'll have to look into that at some point.

doc/config_migration_guide.md
bbaldino: Fixed

README.md
bbaldino: Think it needs to be named `jibri.conf` since it's in `/etc/jtisi/jibri` ?

bgrozev: I find `application` confusing. The only reason we use it is because the undedrlying library refers to it this way. I think most services use similar naming, just scrolling through /etc on a random machine I find `/etc/dhcp/dhclient.conf`, `/etc/ldap/ldap.conf`, `/etc/lvm/lvm.conf`, `/etc/mdadm/mdadm.conf`, `/etc/nginx/nginx.conf`, `/etc/prosody/prosody.cfg.lua`, `/etc/ssl/openssl.cnf`, `/etc/ssh/sshd_config`. Contrast with `/etc/jitsi/videobridge/sip-communicator.properties` :)

bbaldino: `application.conf` actually is a 'special' filename for the typesafe config library: it is loaded automatically as an override to `reference.conf`.  But, since we're using the `-Dconfig.file` argument instead of adding the config dir to the classpath, we can use whatever name we want.  Changed it to `jibri.conf`.

PR #329, opened 2020-08-02
Add support for JWT in webhook requests

bbaldino:


src/main/kotlin/org/jitsi/jibri/util/RefreshingProperty.kt
```diff
@@ -0,0 +1,55 @@
...
+
+package org.jitsi.jibri.util
+
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import kotlin.reflect.KProperty
+
+/**
+ * A property delegate which recreates a value when its accessed after having been
```
JonathanLennox, line 25: "it's"

src/main/resources/reference.conf
```diff
@@ -37,5 +37,24 @@ jibri {
     enable-stats-d = true
   }
   // A list of subscribers interested in receicing webhook events
```
JonathanLennox, line 39: Old typo: "receiving"

src/main/kotlin/org/jitsi/jibri/util/RefreshingProperty.kt
```diff
@@ -0,0 +1,55 @@
...
+    private var valueCreationTimestamp: Instant? = null
+
+    @Synchronized
+    operator fun getValue(thisRef: Any?, property: KProperty<*>): T? {
+        val now = clock.instant()
+        if (valueExpired(now)) {
+            value = try {
+                creationFunc()
+            } catch (t: Throwable) {
+                null
```
JonathanLennox, line 45: You both allow the creationFunc to return null, and catch all its exceptions making the property return null -- do you anticipate both code paths being needed?

src/test/kotlin/org/jitsi/jibri/util/RefreshingPropertyTest.kt
```diff
@@ -0,0 +1,67 @@
...
+import io.kotest.core.spec.style.ShouldSpec
+import io.kotest.matchers.shouldBe
+import io.mockk.spyk
+import org.jitsi.jibri.helpers.FakeClock
+import org.jitsi.jibri.helpers.minutes
+import org.jitsi.jibri.helpers.seconds
+import java.time.Duration
+
+class RefreshingPropertyTest : ShouldSpec({
+    val clock: FakeClock = spyk()
```
JonathanLennox, line 28: `FakeClock` shouldn't need mocking if you take the version we have in jmt.

src/main/kotlin/org/jitsi/jibri/util/RefreshingProperty.kt
bbaldino: I do actually leverage both of those code paths when I use it here.  I think it's basically the semantics of using a `RefreshingProperty`: a value of `null` can't be assumed to be a 'valid' value, it should be considered the lack of an actual value either due to the creation function being unable to produce one or some error occurring.

src/test/kotlin/org/jitsi/jibri/util/RefreshingPropertyTest.kt
bbaldino: The one in JMT doesn't help me much here, unfortunately.  We'll have to move that to `jicoco-test` at some point.

bbaldino: Oh, maybe you meant copy the one from JMT into here.  This is working fine for now, I'll just wait until we get that one in a common place.

PR #331, opened 2020-08-05
Update README.md

samlabs821:
change storage from null to memory, otherwise it gives error:
`Aug 05 11:45:57 internal.auth.domain:muc  error   Error restoring room jvbbrewery@internal.auth.domain from storage: no data storage active`


bbaldino:
An equivalent change was merged in https://github.com/jitsi/jibri/pull/358

PR #335, opened 2020-08-24
feat(config): add timeout override support to the config

zWaR:
Adds new configuration options to the config (old and new configs are
supported) which support overrides for some of the timeout constants Jibri uses internally.


src/main/kotlin/org/jitsi/jibri/config/JibriConfig.kt
```diff
@@ -155,7 +155,13 @@ data class JibriConfig(
     @JsonProperty("finalize_recording_script_path")
     val finalizeRecordingScriptPath: String? = null,
     @JsonProperty("xmpp_environments")
-    val xmppEnvironments: List<XmppEnvironmentConfig>? = null
+    val xmppEnvironments: List<XmppEnvironmentConfig>? = null,
```
bbaldino, line 158: This config should be considered deprecated, so let's not add new properties there, that will just make it harder to phase out.

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/EmptyCallStatusCheck.kt
```diff
@@ -44,6 +46,13 @@ class EmptyCallStatusCheck(
     private fun CallPage.isCallEmpty() = getNumParticipants() <= 1
 
     companion object {
-        val DEFAULT_CALL_EMPTY_TIMEOUT: Duration = Duration.ofSeconds(30)
+        private val defaultCallEmptyTimeout: Duration by config {
+            "JibriConfig::defaultCallEmptyTimeout" {
+                Duration.ofSeconds(Config.legacyConfigSource.defaultCallEmptyTimeout!!)
+            }
+            "jibri.default-call-empty-timeout".from(Config.configSource).convertFrom<Long>(Duration::ofSeconds)
+            "default" { Duration.ofSeconds(30) }
```
bbaldino, line 54: Instead of putting defaults here, you'll want to add them to `reference.conf`

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/EmptyCallStatusCheck.kt
```diff
@@ -44,6 +46,13 @@ class EmptyCallStatusCheck(
     private fun CallPage.isCallEmpty() = getNumParticipants() <= 1
 
     companion object {
-        val DEFAULT_CALL_EMPTY_TIMEOUT: Duration = Duration.ofSeconds(30)
+        private val defaultCallEmptyTimeout: Duration by config {
+            "JibriConfig::defaultCallEmptyTimeout" {
+                Duration.ofSeconds(Config.legacyConfigSource.defaultCallEmptyTimeout!!)
+            }
+            "jibri.default-call-empty-timeout".from(Config.configSource).convertFrom<Long>(Duration::ofSeconds)
```
bbaldino, line 53: Let's just have this parse as a `Duration` directly from `reference.conf`.  Also, let's scope these under a common object.  So you can set this property there like this:
```
jibri {
   call-status-checks {
       call-empty-timeout = 30 seconds
   }   
}
```

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/MediaReceivedStatusCheck.kt
```diff
@@ -52,15 +54,25 @@ class MediaReceivedStatusCheck(
     }
 
     companion object {
+        private val noMediaTimeout: Duration by config {
```
bbaldino, line 57: Same comments as above (no old config, default in `reference.conf` instead of here, make the path `jibri.call-status-checks.no-media-timeout`)

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/MediaReceivedStatusCheck.kt
```diff
@@ -52,15 +54,25 @@ class MediaReceivedStatusCheck(
...
+            "JibriConfig::noMediaTimeout" { Duration.ofSeconds(Config.legacyConfigSource.noMediaTimeout!!) }
+            "jibri.no-media-timeout".from(Config.configSource).convertFrom<Long>(Duration::ofSeconds)
+            "default" { Duration.ofSeconds(30) }
+        }
+        private val allMutedTimeout: Duration by config {
+            "JibriConfig::allMutedTimeout" { Duration.ofMinutes(Config.legacyConfigSource.allMutedTimeout!!) }
+            "jibri.all-muted-timeout".from(Config.configSource).convertFrom<Long>(Duration::ofMinutes)
```
bbaldino, line 64: Same.  `jibri.call-status-checks.all-muted-timeout`.

doc/timeout_overrides.md
```diff
@@ -0,0 +1,69 @@
+# Timeout overrides
+
+Jibri is using several timeouts, which allow it to control and monitor the recording. It is possible to adjust the timeout values and tailor its operations to more specific requirements.
+
+The configuration options are set in the configuration file. Both versions of the configuration file (HOCON and JSON) are supported.
+
+### No Media Timeout
+If all clients have their audio and video muted and if Jibri does not detect any data stream (audio or video) comming in, it will stop recording after `NO_MEDIA_TIMEOUT` expires.
```
bbaldino, line 8: We can get rid of this doc and put this documentation above the fields in `reference.conf`

zWaR:
> Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

FYI, I just signed the CLA.

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/EmptyCallStatusCheck.kt
zWaR: @bbaldino I moved the defaults to `reference.conf`, however the configuration from the `config.conf` does not seem to override it. I am sure I'm doing something wrong ...

bbaldino: Couple things to check:

1. Are you passing the path to `config.conf` to Jibri when starting up? (`-Dconfig.file=/path/to/config.conf`)
1. Did you check that the property names and paths in `config.conf` match `reference.conf` and both of them match what's in the code?

zWaR: Yes on both points, I verified that.

zWaR:
@bbaldino I pushed the changes, but as I mentioned above, the values from the config are always overwritten by `reference.conf`. I will continue looking, but if you find the reason for that, please let me know!

bbaldino:
> @bbaldino I pushed the changes, but as I mentioned above, the values from the config are always overwritten by `reference.conf`. I will continue looking, but if you find the reason for that, please let me know!

I just checked out your branch and overrode `all-muted-timeout` and it worked fine.  What does your config.conf look like?  How are you starting Jibri?

zWaR:
Oh, wow, happy to hear that! I added this at the end of my config:
```HOCON
call-status-checks {
    no-media-timeout = "5 hours"
    all-muted-timeout = "5 hours"
    default-call-empty-timeout = "5 hours"
  }
```

So basically at the end my config looks something like this:
```HOCON
jibri {
  // other options
  call-status-checks {
    no-media-timeout = "5 hours"
    all-muted-timeout = "5 hours"
    default-call-empty-timeout = "5 hours"
  }
}
```

But I must confess that yesterday evening I was reading HOCON specs and as far as I could tell values like `5 hours` should be quoted. I added that to `reference.conf`  as well as to my config file, but afterwards I did not test if that makes any difference.

As for building and testing, I was running a pod in a k8s cluster using `jitsi//base-java:stable-4857` image with an installed `jibri_8.0-34-gd133175-1_all.deb` (from https://download.jitsi.org/stable/) and I was just replacing the jar file with the one I manually built. I was building with `mvn package`. I must say kudos to you, because this was probably the easiest project to build I've ever used. It simply worked, very impressive!

I did modify `launch.sh` though and it looks like this:
```sh
#!/bin/bash
export PATH="/opt/util/:${PATH}"

exec java -Djava.util.logging.config.file=/etc/jitsi/jibri/logging.properties -Dconfig.file=/etc/jitsi/jibri/config.conf -jar /opt/jitsi/jibri/jibri.jar --config "/etc/jitsi/jibri/config.json"
```

If I do not supply the old config with the `--config` flag, I get the following error:

```
root@studio-jibri-75bb6dd48f-2ts9z:/opt/jitsi/jibri# ./launch.sh
2020-08-25 15:03:47.809 INFO: [1] org.jitsi.jibri.Main.handleCommandLineArgs() Jibri run with args []
Exception in thread "main" net.sourceforge.argparse4j.inf.ArgumentParserException: argument -c/--config is required
	at net.sourceforge.argparse4j.internal.ArgumentParserImpl.checkRequiredArgument(ArgumentParserImpl.java:1197)
	at net.sourceforge.argparse4j.internal.ArgumentParserImpl.parseArgs(ArgumentParserImpl.java:881)
	at net.sourceforge.argparse4j.internal.ArgumentParserImpl.parseArgsAtOffsetZero(ArgumentParserImpl.java:709)
	at net.sourceforge.argparse4j.internal.ArgumentParserImpl.parseArgs(ArgumentParserImpl.java:570)
	at net.sourceforge.argparse4j.internal.ArgumentParserImpl.parseArgs(ArgumentParserImpl.java:555)
	at org.jitsi.jibri.MainKt.handleCommandLineArgs(Main.kt:178)
	at org.jitsi.jibri.MainKt.main(Main.kt:52)
```

In order to verify if my config overrides `reference.conf` I focused on `no-media-timeout`. I started a recording and stopped streaming audio and video, but kept the conference participants unmuted. The default timeout for that scenario is 30s, so after 30s conference is closed. With the initially committed code, my configuration from the config changed that interval, but with this last commit, it didn't.

zWaR:
I retested and can confirm that even with quotes the values from the config are overridden by `reference.conf`.

bbaldino:
> I retested and can confirm that even with quotes the values from the config are overridden by `reference.conf`.

Ok, I think we should get to the bottom of this.  In the Jibri main function somewhere, can you add `logger.info(ConfigFactory.load().root().render())` ?  That will dump the loaded typesafe config.

zWaR:
@bbaldino I analyzed everything again and added additional logs to nail it down. Actually, I did something similar to what you are suggesting. I found that the problem was in my environment, in my setup I had a process to sync files up to the pod where I was running the tests and I missed the fact that this was also overwriting my `launch.sh` script, replacing it with one having a wrong filename in the `-Dconfig.file` path. That's why the application was using the defaults from `reference.conf`. I apologize for the confusion.

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/EmptyCallStatusCheck.kt
```diff
@@ -44,6 +46,9 @@ class EmptyCallStatusCheck(
     private fun CallPage.isCallEmpty() = getNumParticipants() <= 1
 
     companion object {
-        val DEFAULT_CALL_EMPTY_TIMEOUT: Duration = Duration.ofSeconds(30)
+        private val defaultCallEmptyTimeout: Duration by config {
+            "jibri.call-status-checks.default-call-empty-timeout".from(Config.configSource)
+        }
+        val DEFAULT_CALL_EMPTY_TIMEOUT: Duration = defaultCallEmptyTimeout
```
bbaldino, line 52: I don't think we need the separate value here, let's get rid of `DEFAULT_CALL_EMPTY_TIMEOUT` and just remove the private from `defaultCallEmptyTimeout`

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/MediaReceivedStatusCheck.kt
```diff
@@ -52,15 +54,21 @@ class MediaReceivedStatusCheck(
...
+        }
+        private val allMutedTimeout: Duration by config {
+            "jibri.call-status-checks.all-muted-timeout".from(Config.configSource)
+        }
         /**
          * How long we'll stay in the call if we're not receiving any incoming media (assuming all participants
          * are not muted)
          */
-        private val NO_MEDIA_TIMEOUT: Duration = Duration.ofSeconds(30)
+        private val NO_MEDIA_TIMEOUT: Duration = noMediaTimeout
```
bbaldino, line 67: Same as above, let's get rid of these and just remove `private` from the others

bbaldino:
Thanks, looks good.  Can you let me know the name on the CLA? You can email me: brian@jitsi.org

PR #337, opened 2020-08-26
change log level for metaconfig to INFO

bbaldino:


lib/logging.properties
```diff
@@ -25,6 +25,7 @@ org.jitsi.jibri.selenium.util.BrowserFileHandler.count = 10
 org.jitsi.jibri.selenium.util.BrowserFileHandler.limit = 10000000
 
 org.jitsi.level = FINE
+org.jitsi.jibri.config.level = INFO
```
bgrozev, line 28: Is this the right package name? I see only `JibriConfig` there and it doesn't even have its own logger

bbaldino: The name we use for the metaconfig logger is manually set to that: https://github.com/jitsi/jibri/blob/master/src/main/kotlin/org/jitsi/jibri/Main.kt#L218

PR #342, opened 2020-09-15
Support non-YouTube RTMP URLs

bbaldino:
This PR introduces a method for supporting non-YouTube RTMP URLs in Jibri.  This is implemented by overloading the existing fields ([streamid](https://github.com/jitsi/jitsi-xmpp-extensions/blob/813e4c453bd3d965ad1b9901bf94f00d3af1badc/src/main/java/org/jitsi/xmpp/extensions/jibri/JibriIq.java#L250) and [you_tube_broadcast_id](https://github.com/jitsi/jitsi-xmpp-extensions/blob/813e4c453bd3d965ad1b9901bf94f00d3af1badc/src/main/java/org/jitsi/xmpp/extensions/jibri/JibriIq.java#L260)) in the following way:

1. If the `streamid` field starts with `rtmp://`, then we treat it as a pre-formed RTMP URL and use it to stream to directly.  Otherwise we assume it's a YouTube stream ID, as before, and create a YouTube RTMP URL from it.

2. If `you_tube_broadcast_id` is set and starts with `http`, then we treat it as a pref-formed _viewing_ URL (a URL at which people can view this stream).  If it is set and doesn't start with `http`, we assume it's a YouTube broadcast ID, as before, and construct a YouTube viewing URL from it.  If it's not set then we do nothing with it.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
+++ b/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
@@ -30,6 +30,7 @@ import org.jitsi.jibri.service.JibriServiceStatusHandler
 import org.jitsi.jibri.service.ServiceParams
 import org.jitsi.jibri.service.impl.SipGatewayServiceParams
 import org.jitsi.jibri.service.impl.StreamingParams
+import org.jitsi.jibri.service.impl.YOUTUBE_URL
 import org.jitsi.jibri.sipgateway.SipClientParams
 import org.jitsi.jibri.status.ComponentState
 import org.jitsi.jibri.status.JibriStatus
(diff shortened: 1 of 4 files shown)


aaronkvanmeerten (dismissed):

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -298,14 +299,30 @@ class XmppApi(
                 )
             }
             JibriMode.STREAM -> {
+                val rtmpUrl = if (startIq.streamId.startsWith("rtmp://", ignoreCase = true)) {
+                    startIq.streamId
+                } else {
+                    "$YOUTUBE_URL/${startIq.streamId}"
+                }
+                val viewingUrl = if (startIq.youtubeBroadcastId != null) {
+                    if (startIq.youtubeBroadcastId.startsWith("http", ignoreCase = true)) {
```
bgrozev, line 308: I think valid youtube broadcast IDs can start with `http`

bbaldino: If so, then we'll just use the URL as-is which I think would be fine.  But previously we always use the `youtubeBroadcastId` field to build a viewing URL via `http://youtu.be/${startIq.youtubeBroadcastId}`.

bgrozev: My point is that if someone's youtube ID happens to be `httpXYZ123` we'll use that as a URL. Instead, we should use `https://youtu.be/httpXYZ123`

bbaldino: Ah, I see.  I intentionally used only `http` to easily match both `http://` and `https://`, but I can explicitly check for either of those instead.

aaronkvanmeerten (approved):
LGTM

PR #344, opened 2020-09-22
use kotest helper to get better errors

bbaldino:


bbaldino:
I think this change is now obsolete.

PR #346, opened 2020-10-06
add extra handling for a call with only jigasi participants so we don…

bbaldino:
…'t fail with no media

We ran into an issue where a call with only jigasi participants would timeout.  Because Jigasi participants don't have any mute state reflected that Jibri can see, this PR adds some special handling to prevent timing out a jibri session when the participants are all jigasi clients.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/status_checks/MediaReceivedStatusCheck.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/status_checks/MediaReceivedStatusCheck.kt
@@ -29,13 +29,18 @@ class MediaReceivedStatusCheck(
         val now = clock.instant()
         val bitrates = callPage.getBitrates()
         // getNumParticipants includes Jibri, so subtract 1
-        val allClientsMuted = callPage.numRemoteParticipantsMuted() == (callPage.getNumParticipants() - 1)
-        logger.info("Jibri client receive bitrates: $bitrates, all clients muted? $allClientsMuted")
+        val numParticipants = callPage.getNumParticipants() - 1
+        val allClientsMuted = callPage.numRemoteParticipantsMuted() == numParticipants
+        val allClientsJigasi = callPage.numRemoteParticipantsJigasi() == numParticipants
+        logger.info("Jibri client receive bitrates: $bitrates, all clients muted? $allClientsMuted, " +
+            "numParticipants: $numParticipants, numJigasis: ${callPage.numRemoteParticipantsJigasi()}")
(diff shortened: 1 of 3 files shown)


JonathanLennox:
Don't you actually want to test `numRemoteParticipantsMuted() + numRemoteParticipantsJigasi() == numParticipants`?

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/MediaReceivedStatusCheck.kt
```diff
@@ -29,13 +29,18 @@ class MediaReceivedStatusCheck(
         val now = clock.instant()
         val bitrates = callPage.getBitrates()
         // getNumParticipants includes Jibri, so subtract 1
-        val allClientsMuted = callPage.numRemoteParticipantsMuted() == (callPage.getNumParticipants() - 1)
-        logger.info("Jibri client receive bitrates: $bitrates, all clients muted? $allClientsMuted")
+        val numParticipants = callPage.getNumParticipants() - 1
+        val allClientsMuted = callPage.numRemoteParticipantsMuted() == numParticipants
+        val allClientsJigasi = callPage.numRemoteParticipantsJigasi() == numParticipants
+        logger.info("Jibri client receive bitrates: $bitrates, all clients muted? $allClientsMuted, " +
+            "numParticipants: $numParticipants, numJigasis: ${callPage.numRemoteParticipantsJigasi()}")
```
damencho, line 36: Do we need to execute js twice, just for the print?

bbaldino:
> Don't you actually want to test `numRemoteParticipantsMuted() + numRemoteParticipantsJigasi() == numParticipants`?

This is slightly different than how I was looking at it, but may be better.  I'll think about it.

src/test/kotlin/org/jitsi/jibri/selenium/status_checks/MediaReceivedStatusCheckTest.kt
```diff
@@ -86,6 +90,13 @@ class MediaReceivedStatusCheckTest : ShouldSpec() {
                     }
                 }
             }
+            context("and all participants are jigasis") {
+                every { callPage.numRemoteParticipantsMuted() } returns 0
+                every { callPage.numRemoteParticipantsJigasi() } returns 3
+                should("not fire any event") {
+                    check.run(callPage) shouldBe null
+                }
+            }
```
JonathanLennox, line 99: Suggestion: add a test case for the some-muted, some-jigasi cases.

JonathanLennox (approved):

PR #355, opened 2020-12-11
Make XMPP server c2s port configurable

amirphl:
Sometimes XMPP c2s port is configured to use a port other than 5222.
For example, I have two XMPP services in my server, one uses 5222 and the other uses 6222.
Jibri always uses port 5222 to connect to XMPP server, therefore, Jibri can't be connected to the second XMPP service which uses port 6222.
In the PR, the port is configured via json config file.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
+++ b/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
@@ -103,10 +103,7 @@ class XmppApi(
                     domain = config.controlLogin.domain
                     username = config.controlLogin.username
                     password = config.controlLogin.password
-                    port = config.controlLogin.port
-
-                    if (port.isNullOrEmpty())
-                        port = "5222"
+                    config.controlLogin.port?.let { port = it.toString() }
 
                     if (config.trustAllXmppCerts) {
                         logger.info("The trustAllXmppCerts config is enabled for this domain, " +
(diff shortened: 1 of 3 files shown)


bbaldino:
Did you sign the CLA?  I think we could probably get this in, but it will need to support falling back to not setting the port in the `MucClientConfiguration` if it's not set in the config object (i.e. fall back to the default).

amirphl:
I just filled the form and signed it then submitted it. Do I have to do anything else?
I updated the code and added the default port.

src/main/kotlin/org/jitsi/jibri/config/JibriConfig.kt
```diff
@@ -30,13 +30,15 @@ import java.io.File
 
 data class XmppCredentials(
     val domain: String = "",
+    val port: String = "",
```
bgrozev, line 33: Can you use an Int please?

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -103,6 +103,10 @@ class XmppApi(
                     domain = config.controlLogin.domain
                     username = config.controlLogin.username
                     password = config.controlLogin.password
+                    port = config.controlLogin.port
+
+                    if (port.isNullOrEmpty())
+                        port = "5222"
```
bbaldino, line 109: @bgrozev any thoughts on handling situations like this?  Wondering if we should define a `default-port` property in `reference.conf` or something to keep defaults more consistent, but I can't remember what we've done in other places.

bgrozev: I don't think we'd ever need to change the default, so adding it to reference.conf doesn't seem appropriate. I would push it to JibiConfig.kt. We already default to 5222 down the line, so I would suggest to only set MucClientConfiguration.port if we have a non-default value

aaronkvanmeerten (approved):

Issue #193, opened 2019-03-31
Cant Connect Jibri Server to Jicofo 


idistech:
Im trying to get an instance of Jibri up and running, and having problems, I think as a result of the jicofo JibriBrewery room not starting...

```
WARNING: [36] org.jitsi.impl.protocol.xmpp.ChatRoomImpl.processPresence().1209 Unable to handle packet: <presence to='focus@auth.video.domain.com/focus201644787312726' from='jibribrewery@internal.auth.video.domain.com/focus' id='3WsmR-21' type='error'><error type='wait'><internal-server-error xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/></error></presence>
Jicofo 2019-03-31 04:26:35.783 SEVERE: [17] org.jitsi.jicofo.xmpp.BaseBrewery.start().196 Failed to create room: JibriBrewery@internal.auth.video.domain.com
net.java.sip.communicator.service.protocol.OperationFailedException: Failed to join the room
```

Ive noticed the service start up timing bug, and tried to restart in the right order, to no change, and cleared out (deleted, twice )  the prosody room through the telnet admin interface, but to no avail.

environment : All 16.04, AWS ( with Generic Kernel on the jibri host ).

Full Logs :
**Jicofo.log**

```
[...]

PR #358, opened 2020-12-12
update Component storage from "null" to "memory"

pjmkrpg:
I propose to update the readme according to the issue #193. For me it fails with the parameter "null" from the readme and does work with "memory".


damencho:
Maybe its worth adding a small comment after that, that it should be none for prosody 0.10 and memory for prosody 0.11.

damencho (approved):

PR #365, opened 2021-01-14
Specify PJSUA as default video input device for sipgw

saghul:
The order of devices is not consistent, so remove any ambiguity by being
explicit.


saghul:
I YOLO MERGED IT FOLKS. NO REGRETS.

PR #372, opened 2021-02-22
Better fix for Process.pidValue jdk11 compatibility

felfert:
Hi,

This PR was inspired by #369 which solves the problem in a sketchy manner:
When attempting to access private fields in java11, a big fat warning is printed which shows, that this is probably a bad idea:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.jitsi.jibri.util.extensions.ProcessKt (file:path-of-jar-with-dependencies.jar) to field java.lang.ProcessImpl.pid
WARNING: Please consider reporting this to the maintainers of org.jitsi.jibri.util.extensions.ProcessKt
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
The proper way to solve the problem ist to refrain from using reflection with newer java versions and instead use conditional compilation to provide a different implementation. Since kotlin does not provide conditional compilation, this PR uses two maven profiles to define different source directories depending on the JDK version used during build.
- With JDK8, the old implementation located below src/main/kotlin-jdk8 is compiled and targets ar set to 1.8
- With JDK>8, the new implementation located below src/main/kotlin-jdk9 is compiled and targets are set to 9

Finally, this PR also adds a small UT for testing the correctness of Process.pidValue

Cheers
 -Fritz

PS: CLA already signed, do you need some proof of that?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,9 +12,6 @@
         <selenium.version>3.12.0</selenium.version>
         <kotlin.version>1.3.71</kotlin.version>
         <exec.mainClass>org.jitsi.jibri.MainKt</exec.mainClass>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
-        <kotlin.compiler.jvmTarget>1.8</kotlin.compiler.jvmTarget>
         <smack.version>4.2.4-47d17fc</smack.version>
         <jersey.version>2.30.1</jersey.version>
         <jackson.version>2.9.5</jackson.version>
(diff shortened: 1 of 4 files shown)


felfert:
> If you haven't already done so, could you please make sure you sign our CLA

Read the last line of my initial comment

felfert:
Forgot to state the obvious:
- Your CI should do multiple builds with different JDK versions for the various distributions
- Of course, there should also be multiple package repositories for the various distributions

BTW: Fedora packaging can be found here (WIP): https://copr.fedorainfracloud.org/coprs/lcts/jitsi/

felfert:
Hold on a moment. There's an even better solution: Creating a[ multi-release jar](https://www.baeldung.com/java-multi-release-jar). I will work on that and come up with another, more elaborated PR. In the meantime, it would be really nice, if you guys could upgrade your CI to use at least JDK9 or greater (preferrably 11)

felfert:
Later...: Well, reading [this](https://blog.gradle.org/mrjars) article, I'm not so shure anymore if a multi-release jar is desireable.
What is your opinion about that?

felfert:
Anyone cares to review and/or merge?

bbaldino:
I'm not sure if I'll end up merging this or not.  We only officially support Java 8 right now and at some point will probably just make the transition to supporting Java 11 and not the older versions.  The current Java 11 support "hack" was minor enough that I was ok to just merge it in.  I'll think about this, though.

damencho:
Fixed in https://github.com/jitsi/jibri/pull/517

PR #376, opened 2021-03-15
A few test fixes, upgrade kotest to 4.3.2

bbaldino:
While starting to look into a flaky unit test I found some other test issues, so wanted to address those first.  These changes fix a couple minor things, upgrade kotest to 4.3.2 and add a junit xml reporter which uses the full tree of kotest scopes as the test name to help reduce ambiguity when a test fails.


JonathanLennox (commented):
LGTM assuming tests pass

PR #380, opened 2021-03-23
Upgrade kotlin version from 1.3.71 to 1.4.31

bbaldino:


bbaldino:
I'll fix the `jtor` -> `ktor` typo in the commit when merging.

src/main/kotlin/org/jitsi/jibri/webhooks/v1/WebhookClient.kt
```diff
@@ -83,6 +83,7 @@ class WebhookClient(
     }
 
     private val client = client.config {
+        expectSuccess = false
```
JonathanLennox, line 86: Is this related to the Kotlin change?

bbaldino: It was something I encountered from the ktor upgraded, I explained it in that commit message (easy to miss in github):
```
note that we now need to explicitly set "expectSuccess" to false to
disable the default response validator in the ktor client.  this was
always the desired behavior, but due to a bug
(https://youtrack.jetbrains.com/issue/KTOR-1412) the response validation
was not kicking in before.
```

JonathanLennox (approved):

PR #383, opened 2021-03-24
Logger improvements

bbaldino:
This PR

- Uses the jitsi-utils logger instead of java.util.logging directly, to be consistent with the other jitsi apps
- Implements a logger hierarchy for services and their helper classes to give better log context


src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/FfmpegCapturer.kt
```diff
@@ -83,14 +85,14 @@ class FfmpegCapturer(
...
-        logger.debug("Detected os as OS: $osType")
+        logger.debug { "Detected os as OS: $osType" }
         getCommand = when (osType) {
             OsType.MAC -> { sink: Sink -> getFfmpegCommandMac(FfmpegExecutorParams(), sink) }
             OsType.LINUX -> { sink: Sink -> getFfmpegCommandLinux(FfmpegExecutorParams(), sink) }
             else -> throw UnsupportedOsException()
         }
 
-        ffmpeg.addStatusHandler(this::onFfmpegProcessUpdate)
+        this.ffmpeg.addStatusHandler(this::onFfmpegProcessUpdate)
```
bbaldino, line 95: The `this` here (and a few other places) is necessary because otherwise `ffmpeg` resolves to the one passed in to the constructor, which is nullable.  The other alternative is to make the ctor arg something like `_ffmpeg`, but I find that awkward when calling using named parameters, so went with this approach.

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -98,11 +97,16 @@ data class RecordingMetadata(
...
     private val fileRecordingParams: FileRecordingParams,
-    private val jibriSelenium: JibriSelenium = JibriSelenium(),
-    private val capturer: FfmpegCapturer = FfmpegCapturer(),
+    jibriSelenium: JibriSelenium? = null,
+    capturer: FfmpegCapturer? = null,
     private val processFactory: ProcessFactory = ProcessFactory(),
     fileSystem: FileSystem = FileSystems.getDefault()
 ) : StatefulJibriService("File recording") {
+    init {
+        logger.addContext("session_id", fileRecordingParams.sessionId)
```
bgrozev, line 106: Should this have it's own logger rather than reusing super's? It would also avoid multiple "init" blocks.

bgrozev (approved):

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
bbaldino: I don't think we usually use separate loggers for parent classes (just "enclosing" classes) since the classname would end up being the same.

PR #390, opened 2021-04-02
Handle a recording which was stopped before any media was recorded

bbaldino:
Sometimes a Jibri gets a stop request before it's even started up ffmpeg.  Currently, that results in Jibri writing a metadata file to the recording directory (and invoking the finalize script) even though there isn't actually any recording.  This doesn't make much sense, and can be confusing/noisy when looking through failed recording directories.

This PR detects whether or not a media file has been written in `FileRecordingJibriService#stop` and, if no file has been written, Jibri deletes the recording directory and skips calling the finalize script.


bbaldino:
I was able to simulate the behavior pretty well in the unit test, but I've not yet had the chance to try it on a real Jibri yet.  I think we should validate both the happy path (making sure when a recording is written things still work) and the 'no recording file' path (making sure the directory gets deleted and finalize isn't called).

raluca8x8 (approved):

PR #398, opened 2021-04-18
Adding XMPP Port option to config file

kpeiruza:
It allows Jibri to be configured with a non-standard XMPP port.

It's handy when you have Prosody using non standard ports (behind a loadbalancer/firewall, several deployments using the same IP, etc...)

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
+++ b/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
@@ -111,6 +111,14 @@ class XmppApi(
                         disableCertificateVerification = config.trustAllXmppCerts
                     }
 
+                    // FLOSS PATCH, 2021-01-12
+                    if (config.xmppPort != "") {
+                        logger.info("The xmppPort config is enabled for this Jibri")
+                        port = config.xmppPort
+                    } else {
+                        port = "5222"
+                    }
+
                     val recordingMucJid =
                         JidCreate.bareFrom("${config.controlMuc.roomName}@${config.controlMuc.domain}").toString()
                     val sipMucJid: String? = config.sipControlMuc?.let {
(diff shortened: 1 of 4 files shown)


kpeiruza:
I've just signed CCLA.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -111,6 +111,14 @@ class XmppApi(
                         disableCertificateVerification = config.trustAllXmppCerts
                     }
 
+                    // FLOSS PATCH, 2021-01-12
```
bgrozev, line 114: A couple of comments here:
1. Please make the config property an `Int` (I know MucClientConfiguration expects a string, but that's awkward, so we can keep is isolated there).
2. Either include the port number in the log message, or remove the message (I'm prety sure it will get printed elsewhere)
3. Remove the "FLOSS PATCH" comment. 

It could be simplified to just `port = config.xmppPort.toString()`

bgrozev:
Thanks for your contribution! I left a comment, but otherwise it looks good. 

I can see the CLA, please ignore Jenkins if it asks again.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
kpeiruza: Hi, I've tried to change it to Int but then a bunch of errors arise... sorry, I'm a Sysadmin and I can code in some langs but Kotlin isn't one of them.

If you can change it easily, please, feel free to go ahead, otherwise I'll check again on another bank holiday... right now I'm dealing with this to build a new docker image supporting Jibri+Pulseaudio+xmppPort :'(

Thanks!

PS: OFC, feel free to drop any comments as well :-)

bgrozev:
This looks like a dup of #355

bgrozev:
Closing in favor of #355 which was opened first (and we'll merge shortly).

PR #401, opened 2021-04-27
Stop using email for sip jibri callstatsusername, but instead use a c…

raluca8x8:
…ustom value

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
@@ -64,13 +64,20 @@ data class CallParams(
      * This will be used by jitsi-meet to send the conference passcode to prosody
      * and bypass the password prompt
      */
-    val passcode: String? = null
+    val passcode: String? = null,
+    /**
+     * Override the default value of callStats username.
+     * Note that this is currently only used in the sipgateway gateway scenario;
+     */
+    val callStatsUsername: String = ""
 ) {
     override fun toString(): String {
         return if (passcode.isNullOrEmpty()) {
-            "CallParams(callUrlInfo=$callUrlInfo, email='$email', passcode=$passcode)"
+            "CallParams(callUrlInfo=$callUrlInfo, email='$email', passcode=$passcode" +
+                    ", callStatsUsername=$callStatsUsername)"
         } else {
-            "CallParams(callUrlInfo=$callUrlInfo, email='$email', passcode=*****)"
+            "CallParams(callUrlInfo=$callUrlInfo, email='$email', passcode=*****" +
+                    ", callStatsUsername=$callStatsUsername)"
         }
     }
 }
@@ -96,6 +103,11 @@ data class JibriSeleniumOptions(
      */
     val email: String = "",
     /**
+     * The callstats username to be used for jibri.
+     * Set this only to override the default callStatsUsername
+     */
+    val callStatsUsername: String = "",
+    /**
      * Chrome command line flags to add (in addition to the common
      * ones)
      */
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -64,13 +64,20 @@ data class CallParams(
...
-    val passcode: String? = null
+    val passcode: String? = null,
+    /**
+     * Override the default value of callStats username.
+     * Note that this is currently only used in the sipgateway gateway scenario;
+     */
+    val callStatsUsername: String = ""
 ) {
     override fun toString(): String {
         return if (passcode.isNullOrEmpty()) {
```
bbaldino, line 75: I'm wondering if, rather than using a custom `toString` to handle masking of the passcode (which then needs to be updated for every field) we should just do something like:

```
override fun toString(): String = this.copy(passcode = "*******")
```

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -259,10 +272,17 @@ class JibriSelenium(
             try {
                 HomePage(chromeDriver).visit(callUrlInfo.baseUrl)
 
+                var callStatsUsername = "jibri"
+                if (jibriSeleniumOptions.callStatsUsername.isNotEmpty()) {
+                    callStatsUsername = jibriSeleniumOptions.callStatsUsername
```
bbaldino, line 277: Would it be best to still prefix this with something like `jibri-sip` ?

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -259,10 +272,17 @@ class JibriSelenium(
...
+                } else if (MainConfig.jibriId.isNotEmpty()) {
+                    callStatsUsername = MainConfig.jibriId
+                }
```
bbaldino, line 280: You lost the case where it defaults to `"jibri"` if the `jibriId` is empty.

raluca8x8: This case is covered by initializing callStatsUsername = "jibri" at the beginning, and only overriding it with jibriId if callStatsUsername is empty and if jibriId is not empty

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
raluca8x8: I prefer letting the entity calling the start sip gateway service to decide how the callStatsUsername will be set. And that entity can decide to add sip-jibri prefix to the callstats username.

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
raluca8x8: I'll give it a try, sounds promising

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
bbaldino: 🤦  Missed that. 👍

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
bbaldino: Ok.  Maybe we want to call the `JibriSeleniumOptions` field `callStatsUsernameOverride`?

raluca8x8: Renamed it now, it does sound clearer this way

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
raluca8x8: Unfortunately this did not work, it resulted in stack overflow, as it keeps copying and calling toString on the copy.

bbaldino: Ah damn, ok.

bbaldino (approved):

oanaianc (approved):

aaronkvanmeerten (approved):

PR #404, opened 2021-05-05
Use a proxy for all the outbound SIP calls

oanaianc:


bbaldino:
jenkins, add to whitelist

raluca8x8 (approved):

PR #406, opened 2021-05-12
Add the option to configure the base-url of the call used by Jibri to…

oanaianc:
… join

### DIFF
--- a/doc/example_xmpp_envs.conf
+++ b/doc/example_xmpp_envs.conf
@@ -11,6 +11,9 @@
             // The base XMPP domain
             xmpp-domain = "xmpp-domain"
 
+            // An (optional) base url the Jibri will join if it is set
+            // base-url = "https://meet.example.com"
+
             // The MUC we'll join to announce our presence for
             // recording and streaming services
             control-muc {
--- a/src/main/kotlin/org/jitsi/jibri/config/JibriConfig.kt
+++ b/src/main/kotlin/org/jitsi/jibri/config/JibriConfig.kt
@@ -72,6 +72,11 @@ data class XmppEnvironmentConfig(
     @JsonProperty("xmpp_domain")
     val xmppDomain: String,
     /**
+     * The baseUrl we'll use, if set, to join the call
+     */
+    @JsonProperty("baseUrl")
+    val baseUrl: String?,
+    /**
      * The login information for the control API
      */
     @JsonProperty("control_login")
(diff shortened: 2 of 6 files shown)


raluca8x8 (dismissed):

doc/example_xmpp_envs.conf
```diff
@@ -11,6 +11,9 @@
             // The base XMPP domain
             xmpp-domain = "xmpp-domain"
 
+            // An (optional) base url the Jibri will join if it is set
+            base-url = "http://meet.jit.si"
```
saghul, line 15: Can you please leave it commented out so people don't enable it by mistake?

Also use https://meet.example.com to make it obvious it needs changing.

saghul (commented):
This is awesome, thanks for the quick turnaround! Left a minor comment.

doc/example_xmpp_envs.conf
oanaianc: done

raluca8x8 (approved):

saghul (approved):

src/main/kotlin/org/jitsi/jibri/config/JibriConfig.kt
```diff
@@ -71,6 +71,11 @@ data class XmppEnvironmentConfig(
      */
     @JsonProperty("xmpp_domain")
     val xmppDomain: String,
+    /**
+     * The baseUrl we'll use, if set, to join the call
+     */
+    @JsonProperty("baseUrl")
+    val baseUrl: String?,
```
bbaldino, line 78: We shouldn't add new properties here, as this config is deprecated

bbaldino:
Sorry for the late comment, I was out of town, but it's just a minor one anyway.  Glad this ended up being able to be implemented so easily (I thought it was going to require Jicofo changes).

saghul:
Thanks for checking anyway! This helped get Docker working nicely indeed!

src/main/kotlin/org/jitsi/jibri/config/JibriConfig.kt
```diff
@@ -122,6 +127,9 @@ fun com.typesafe.config.Config.toXmppEnvironment(): XmppEnvironmentConfig =
         name = getString("name"),
         xmppServerHosts = getStringList("xmpp-server-hosts"),
         xmppDomain = getString("xmpp-domain"),
+        baseUrl = if (hasPath("base-url")) {
```
Tinkesh-Kumar, line 130: Can we allow this to set at run time ?, Similar to HttpAPI.

PR #414, opened 2021-06-14
Allow sip clients to specify the sip schema or use `sip` as the default value

oanaianc:


raluca8x8 (approved):

src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
```diff
@@ -40,6 +43,9 @@ class PjsuaClient(
 ) : SipClient() {
     private val logger = createChildLogger(parentLogger)
     private val pjsua: JibriSubprocess = JibriSubprocess(logger, "pjsua")
+    private val sipOutboundPrefix: String? by optionalconfig(
```
bbaldino, line 46: Instead of making this optional and putting the default in the code, I'd make it non-optional and put the default value of `sip` in the default config file.

bbaldino: Oops, sorry I misread this...I see this isn't the scheme.

src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
```diff
@@ -70,14 +76,31 @@ class PjsuaClient(
...
 
+        /**
+         * A SIP address is written in user@domain.tld format in a similar fashion to an email address.
+         * An address like: sip:1-999-123-4567@voip-provider.example.net
+         *
+         * The default sip scheme is `sip` if none is specified; Valid options would be `sip` or `sips`
+         */
+        var sipScheme = "sip"
+        var sipAddress = pjsuaClientParams.sipClientParams.sipAddress
+        if (sipAddress.isNotEmpty() && sipAddress.hasSipSchemeEmbedded()) {
```
bbaldino, line 87: A small nit, but I might do this as:

create a `String.getSipScheme(): String?` that returns either the scheme or null, and then you can do:

`val sipScheme = pjsuaCLientParams.sipClientParams.getSipScheme() ?: "sip"`

And then also create a `String.getSipAddress(): String` that checks if the scheme is there and does the appropriate thing so then you get.

`val sipAddress = pjsuaCLientParams.sipClientParams.sipAddress.getSipAddress()`

Not a huge deal, but just a thought (it gets rid of the vars and the needs for assigning twice).

src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
```diff
@@ -92,14 +115,22 @@ class PjsuaClient(
...
             if (pjsuaClientParams.sipClientParams.userName != null) {
-                command.add("--proxy=sip:${pjsuaClientParams.sipClientParams.userName.substringAfter('@')};" +
-                        "transport=tcp;hide")
+                command.add(
+                    "--proxy=$sipScheme:${pjsuaClientParams.sipClientParams.userName.substringAfter('@')};" +
+                            "transport=tcp;hide"
+                )
+            }
+            if (!sipOutboundPrefix.isNullOrEmpty()) {
+                sipAddress = sipOutboundPrefix.plus(sipAddress)
```
bbaldino, line 124: Ah, I see this is a bit tricky for sipAddress since it has to also have the outbound prefix appended, you could do something like:

```
if (sipOutboundPrefix.isNUllOrEmpty() {
    command.add("$sipScheme:$sipAddress$sipOutboundPrefix")
} else {
    command.add("$sipScheme:$sipAddress")
}
```

bbaldino (dismissed):
lgtm, had some ideas on a few tweaks but not critical.

oanaianc:
Thank you @bbaldino for the suggestions, I updated the code

raluca8x8 (approved):

PR #417, opened 2021-06-23
Exit fast when sip jibri participant is kicked out

raluca8x8:


src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -212,10 +218,14 @@ class JibriSelenium(
         // Note that the order here is important: we always want to check for no participants before we check
         // for media being received, since the call being empty will also mean Jibri is not receiving media but should
         // not cause Jibri to go unhealthy (like not receiving media when there are others in the call will).
-        val callStatusChecks = listOf(
+        val callStatusChecks = mutableListOf(
```
bbaldino, line 221: Instead of using `mutableListOf`, I think you can use `buildList` which will let you embed conditional logic into the block but still keep the list immutable.

src/main/kotlin/org/jitsi/jibri/selenium/status_checks/LocalParticipantKickedStatusCheck.kt
```diff
@@ -0,0 +1,40 @@
...
+) : CallStatusCheck {
+    private val logger = createChildLogger(parentLogger)
+
+    init {
+        logger.info("Starting local participant kicked out call check")
+    }
+
+    // The timestamp at which we last saw the call transition from
+    // joined to kicked out
+    private val localParticipantKickedTime = StateTransitionTimeTracker(clock)
```
bbaldino, line 22: I don't think you need/want `StateTransitionTimeTracker` here.  You don't want to wait for some amount of time after it was kicked to leave, right?  You just want to finish as soon as it was kicked?  I think you just want:
```
return if (callPage.isLocalParticipantKicked()) {
    logger.info(...)
    SeleniumEvent.LocalParticipantKicked
} else {
    null
}
```

raluca8x8: You are right, I will probably not need this. I applied your suggestion

src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
raluca8x8: Thank you, done!

bbaldino (approved):

Issue #429, opened 2021-08-08
Allow multiple jibri baseurl for single Xmpp host


Tinkesh-Kumar:
If multiple meeting host is using single prosody, Jibri will not allow to record the session as base url will be same for both company1 & company2.
Ex:-
meet.company1.com is using XMPP host1
meet.company2.com is also using XMPP host1

In this case, since jibri allow only one base url for XMPP host, Recording will not work for both company1 & company2. Is my understanding correct?

Is there any reason for single baseurl for xmpp hosts?

oanaianc:
If you want to setup both company1 & company2, you will have to set two xmpp environments, one for each domain. 
Also the baseUrl is not mandatory; if is set the jibri will join it, otherwise it will join the https://xmpp-domain
eg. 
xmpp {
environments = [
                {
                    name = "company1",
                    xmpp-server-hosts = [""],
                    xmpp-domain = "meet.company1.com",
                    base-url = "",
                    control-login {
          ......
                },
                {
                    name = "company2",
                    xmpp-server-hosts = [""],
                    xmpp-domain = "meet.company2.com",
                    base-url = "",
                    control-login {
          ......
                },
]
}

Tinkesh-Kumar:
We wanted to optimise this. We do not want to create multiple xmpp host just to support jibri. I see httpAPI already support this. Can same thing being done for XMPP API.

aaronkvanmeerten:
The main reasoning for not doing this in the XMPP message was to avoid jibri being launched against unsupported domains, via a maliciously built XMPP message.  So, one way to handle this will be to have a list of allowed domains in jibri, and check the incoming requested domain against that.  I believe the team would be open to pull requests that implemented this functionality, if anyone is interested.  Changes will likely be needed in Jicofo and Jibri to fully enable this functionality.

aaronkvanmeerten:
Another thought I had: if XMPP host1 is being re-used, then either URL will work, no?  Unless you have special prosody configuration, two webservers with meet.company1.com and meet.company2.com pointed to XMPP host1 will simply re-use the same prosody backend.  This means that meet.company1.com/abc and meet.company2.com/abc are equivalent, unless you've done some special configuration of both config.js for meet, and prosody.cfg.lua to provide separate MUCs for both.  Once you've already done this much customization, one additional jibri configuration doesn't seem particularly onerous to support this use-case.

Tinkesh-Kumar:
[...]

PR #430, opened 2021-08-09
Updated jitsi-xmpp-extensions version & send jibri base url in xmppcl…

Tinkesh-Kumar:
…ient

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
+++ b/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
@@ -285,7 +285,8 @@ class XmppApi(
         val appData = startIq.appData?.let {
             jacksonObjectMapper().readValue<AppData>(startIq.appData)
         }
-        var baseUrl = if (appData?.baseUrl == null || appData.baseUrl.isEmpty() ) xmppEnvironment.baseUrl else appData.baseUrl;
+        var baseUrl = if (appData?.baseUrl == null || appData.baseUrl.isEmpty())
+            xmppEnvironment.baseUrl else appData.baseUrl
         val callUrlInfo = getCallUrlInfoFromJid(
             startIq.room,
             xmppEnvironment.stripFromRoomDomain,
--- a/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
+++ b/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
@@ -286,7 +286,7 @@ class XmppApi(
             startIq.room,
             xmppEnvironment.stripFromRoomDomain,
             xmppEnvironment.xmppDomain,
-            xmppEnvironment.baseUrl
+            startIq.baseUrl
         )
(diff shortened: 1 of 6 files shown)


Tinkesh-Kumar:
Xmpp-extension upgrade PR.
https://github.com/jitsi/jitsi-xmpp-extensions/pull/44/files

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -286,7 +286,7 @@ class XmppApi(
             startIq.room,
             xmppEnvironment.stripFromRoomDomain,
             xmppEnvironment.xmppDomain,
-            xmppEnvironment.baseUrl
+            startIq.baseUrl
```
aaronkvanmeerten, line 289: My concern is that we're not doing any checking of this baseUrl.  Without addition checks, this could be used by a malicious user to send your jibris to a poisoned site which records the credentials jibri uses when logging in.  In addition, this entirely overrides the Xmpp API without keeping the previous behavior as default.  In my opinion this should be a separate section of code which first checks for a startIq baseUrl and confirms it's allowed, and then uses the environment base URL as a backup the start IQ doesn't include it.

aaronkvanmeerten (changes requested):

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
Tinkesh-Kumar: We do not do any validation for baseUrl in case of Http, 
https://github.com/jitsi/jibri/blob/master/doc/http_api.md

saghul:
I'm -1 on this approach. Jibri is already able to connect to several XMPP environments and having one extra connection should not be problematic since it uses very little traffic.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
aaronkvanmeerten: This is because the HTTP api is entirely a different use case.  It's meant to be for jibris who expose a command and control interface.  It's up to whomever is building that to control access to it.  In the case of XMPP it's a much more exposed interface and therefore deserves this level of checks.  Without this check, every default jibri and jitsi-meet install will be vulnerable to this type of attack.

Tinkesh-Kumar:
> I'm -1 on this approach. Jibri is already able to connect to several XMPP environments and having one extra connection should not be problematic since it uses very little traffic.

@aaronkvanmeerten Thanks for reviewing. Are you asking to keep an allowed list of baseUrl in xmpp jibri config & validate the same here.
 @saghul Thanks for reviewing. Here is the requirement https://github.com/jitsi/jibri/issues/429#issuecomment-895206741. Here concern is not about the xmpp connection. We wanted to allow multiple jibri url for single xmpp host.

saghul:
You can have a jibri connect to the same host under 2 different names, if you have DNS records for it.

PR #448, opened 2021-11-02
chore: Fix ktlint errors.

bgrozev:


JonathanLennox (approved):
Do we not have the code to check these automatically?

bgrozev:
> Do we not have the code to check these automatically?

I suspect it's running with a different version of ktlint (since we still use maven-antrun-plugin in jibri), because I got a disjoint set of errors locally and on GH. The tests passed, so this change satisfies both.

PR #468, opened 2022-03-03
Update smack, remove "half-open" detection.

bgrozev:
- fix: Update Smack to 4.4.5, remove "half-open" detection.
- feat: Add a metric when recording is stopped due to XMPP disconnecting.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
+++ b/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
@@ -33,6 +33,7 @@ import org.jitsi.jibri.service.impl.StreamingParams
 import org.jitsi.jibri.service.impl.YOUTUBE_URL
 import org.jitsi.jibri.sipgateway.SipClientParams
 import org.jitsi.jibri.statsd.JibriStatsDClient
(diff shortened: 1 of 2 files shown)


damencho (approved):

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -127,6 +120,7 @@ class XmppApi(
             val environmentContext = createEnvironmentContext(xmppEnvironment, mucClient)
             if (jibriManager.currentEnvironmentContext == environmentContext) {
                 logger.warn("XMPP disconnected, stopping.")
+                statsDClient?.incrementCounter(STOPPED_ON_XMPP_CLOSED, mucClient.tags())
```
raluca8x8, line 123: In the comment for STOPPED_ON_XMPP_CLOSED you mention "A recording session was stopped because XMPP disconnected.". However, right now, this metric only records when Jibri stops, it does not take into account whether a recording is in progress at that time.

Therefore, you should probably first check if there is currentActiveService in jibriManager and only if so increment the STOPPED_ON_XMPP_CLOSED counter.

PR #474, opened 2022-04-04
fix: Do not expect ICE connected when there are no participants.

bgrozev:


raluca8x8 (approved):

bbaldino:
> > For Jibri to be in the call, it must be connected via the bridge (we disable p2p when Jibri joins, as far as I can remember)...so shouldn't ICE always be connected, even if Jibri is the only participant?
> 
> Jicofo only initiates sessions when there are at least 2 participants (at least in our environments).

Oh!  Ok, I think I follow now (previously there was no way to bring Jibri into a call without some participant already in the call bringing it in).

raluca8x8:
> > > For Jibri to be in the call, it must be connected via the bridge (we disable p2p when Jibri joins, as far as I can remember)...so shouldn't ICE always be connected, even if Jibri is the only participant?
> > 
> > 
> > Jicofo only initiates sessions when there are at least 2 participants (at least in our environments).
> 
> Oh! Ok, I think I follow now (previously there was no way to bring Jibri into a call without some participant already in the call bringing it in).

This change is important especially for the SIP-Jibri use case, as a SIP participant can be present alone in the meeting for a while (for now we allow 30 minutes).

PR #479, opened 2022-05-03
feature: allow port in xmppServerHosts

aaronkvanmeerten:


JonathanLennox:
Is it possible to add a test to XmppApiTest for this case?

aaronkvanmeerten:
> Is it possible to add a test to XmppApiTest for this case?

I have added at least an example of both methods, although the confirmation of the value being set was outside my jicofo-fu

JonathanLennox (approved):

PR #493, opened 2022-10-17
fix: Authorization header sending for webhooks

cristianbotiza:
- the issue was caused by io.jsonwebtoken having a hardcoded call to javax.xml.bind.* packages; these are gone in java 9+. The fix was to simply upgrade the jsonwebtoken library. Use latest, 0.11.5. This also requires a split (api vs impl vs jackson), so add the API as compile dependency and the impl and jackson serializer as runtime dependencies
 - add logging around RefreshingProperty throwing exceptions
 - separate Throwable (unrecoverable, propagate it) vs Exception (return null, rely on next refresh, but log)
 - update unit tests
 - use bearerAuth() instead of header(), to reduce verbosity

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,7 @@
         <kotest.version>5.3.0</kotest.version>
         <mockk.version>1.10.0</mockk.version>
         <ktor.version>2.0.2-jitsi</ktor.version>
+        <jwt.version>0.11.5</jwt.version>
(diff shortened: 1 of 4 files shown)


raluca8x8 (approved):

oanaianc (approved):

raluca8x8:
> `1 check was pending ` 🔥

Did not see that, I will keep an eye on the result then

cristianbotiza:
what check, Saul?

Issue #496, opened 2022-12-05
Hard-coded --id argument breaks the video SIP call if it is triggered from UI (SIP-Jibri related)


emrahcom:
## Description
When I try to start a video SIP session by inviting the other peer from UI, `jibri` adds an `--id` argument with a hard-coded value:
`--id=$displayName <sip:jibri@127.0.0.1>`

This value overwrites the `--id` value which comes from `/etc/jitsi/jibri/pjsua.config` and the hard-coded value doesn't work for my system.

## Current behavior
The SIP session fails because the hard-coded `--id` doesn't match my environment.

## Expected Behavior
`jibri` should not add a had-coded `--id` and allow admin to set it in `/etc/jitsi/jibri/pjsua.config` according to their environment.

## Possible Solution
Removing the following codes (`else` case):

https://github.com/jitsi/jibri/blob/master/src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt#L99-L100

## Steps to reproduce
Invite SIP peer from UI through `SIP-jibri`. Maybe this doesn't fail for all system if `jibri@127.0.0.1` has a meaning for its SIP server.

# Environment details

- Debian 11 Bullseye
- Jibri 8.0-140-gccc7278-1
- openjdk-11-jre-headless 11.0.16+8-1~deb11u1
- AMD64

saghul:
What kind of value do you want to set there?

emrahcom:
For example `--id "1009 <sip:1009@freeswitch.mydomain.corp>"` which works for my SIP server.

`sip:1009@127.0.0.1` or `sip:jibri@127.0.0.1` fails for me.

emrahcom:
These are what I see on FreeSWITCH logs:

```
2022-12-05 19:28:34.430967 99.17% [INFO] sofia.c:10462 sofia/internal/jibri@127.0.0.1 receiving invite from ...                                                                                                                        
nua_stack.c:599 nua_stack_signal() nua((nil)): recv signal r_unref                                                                                                            
2022-12-05 19:28:34.430967 99.17% [DEBUG] sofia.c:10556 verifying acl "domains" for ip/port 172.17.17.203:0.                                                                  
2022-12-05 19:28:34.430967 99.17% [WARNING] sofia_reg.c:3223 Can't find user [1009@127.0.0.1] from 172.17.17.203                                                              
You must define a domain called '127.0.0.1' in your directory and add a user with the id="1009" attribute                                                                     
[...]

PR #502, opened 2023-03-07
refactor: remove hardcoded SIP id

emrahcom:
This commit fixes the issue #496

It removes the hardcoded `--id` (which is `<sip:jibri@127.0.0.1>`) from `Jibri` and puts it into [pjsua.config](https://github.com/jitsi/jibri/blob/master/resources/debian-package/etc/jitsi/jibri/pjsua.config)


saghul (approved):
LGTM. @oanaianc can you PTAL too?

damencho (approved):

damencho:
We had already discussed ot with @oanaianc

PR #505, opened 2023-03-08
feat: add the optional proxy parameter for pjsua

emrahcom:
This commit adds an optional `--proxy` parameter for `pjsua` client. It allows to set  SIP `proxy` through `jitsi-component-selector` API.

If there is no `proxy` in the API request, the application still sets the forced proxy to not break the old setups.
`--proxy=...transport=tcp;hide`

A sample request may be like the following

```json
{
  "callParams": {
    "callUrlInfo": {
      "baseUrl": "$JITSI_HOST",
      "callName": "$JITSI_ROOM"
    }
  },
  "componentParams": {
    "type": "SIP-JIBRI",
    "region": "default-region",
    "environment": "default-env"
  },
  "metadata": {
    "sipClientParams": {
      "userName": "$CALLER_USERNAME",
      "password": "$CALLER_PASSWORD",
      "sipAddress": "$CALLEE",
      "displayName": "Caller",
      "proxy": "$SIP_PROXY",
      "autoAnswer": false
    }
  },
  "callLoginParams": {
    "domain": "sip.jitsi.mydomain.com",
    "username": "sip",
    "password": "$PROSODY_SIP_PASSWD"
  }
}
```

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
+++ b/src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
@@ -98,7 +98,7 @@ class PjsuaClient(
             command.add("--password=${pjsuaClientParams.sipClientParams.password}")
         }
 
-        if (pjsuaClientParams.sipClientParams.contact != null) {
+        if (!pjsuaClientParams.sipClientParams.contact.isNullOrEmpty()) {
             command.add("--contact=${pjsuaClientParams.sipClientParams.contact}")
         }
 
--- a/src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
+++ b/src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
@@ -102,13 +102,19 @@ class PjsuaClient(
             command.add("--contact=${pjsuaClientParams.sipClientParams.contact}")
         }
 
+        if (!pjsuaClientParams.sipClientParams.proxy.isNullOrEmpty()) {
+            command.add("--proxy=${pjsuaClientParams.sipClientParams.proxy}")
+        }
+
         if (pjsuaClientParams.sipClientParams.autoAnswer) {
             command.add("--auto-answer-timer=30")
             command.add("--auto-answer=200")
         } else {
             // The proxy we'll use for all the outgoing SIP requests;
+            // This proxy will be enabled if --proxy is not set explicitly through API
             // The client should not specify a Route header in the sip INVITE message. Using hide will let the server set the Route header
-            if (pjsuaClientParams.sipClientParams.userName != null) {
+            if (pjsuaClientParams.sipClientParams.proxy == null &&
+                pjsuaClientParams.sipClientParams.userName != null) {
                 command.add(
                     "--proxy=$sipScheme:${pjsuaClientParams.sipClientParams.userName.substringAfter('@')};" +
                         "transport=tcp;hide"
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
```diff
@@ -98,17 +98,24 @@ class PjsuaClient(
             command.add("--password=${pjsuaClientParams.sipClientParams.password}")
         }
 
-        if (pjsuaClientParams.sipClientParams.contact != null) {
+        if (!pjsuaClientParams.sipClientParams.contact.isNullOrEmpty()) {
```
damencho, line 101: Is it possible pjsuaClientParams.sipClientParams.contact is null?

src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
```diff
@@ -98,17 +98,24 @@ class PjsuaClient(
...
             command.add("--contact=${pjsuaClientParams.sipClientParams.contact}")
         }
 
+        if (!pjsuaClientParams.sipClientParams.proxy.isNullOrEmpty()) {
```
damencho, line 105: Same here, is it possible that pjsuaClientParams.sipClientParams.proxy is null?

src/main/kotlin/org/jitsi/jibri/sipgateway/pjsua/PjsuaClient.kt
emrahcom: Yes, if it is not set in the JSON request, its default value is `null` which comes from [SipClient.kt](https://github.com/jitsi/jibri/blob/master/src/main/kotlin/org/jitsi/jibri/sipgateway/SipClient.kt#L23-L53)

damencho: Then that will produce a NPE?

emrahcom: I didn't understand what you meant. They 3 possibilities:

- The `contact` is not in JSON which comes from API request. In this case, it is `null`
- The `contact` is in JSON but its value is an empty string, then it is `""` in the application to
- The `contact` is in JSON and its value is a string (not empty)

damencho: If pjsuaClientParams.sipClientParams.contact is null, then doing `pjsuaClientParams.sipClientParams.contact.isNullOrEmpty()` in kotlin is ok and not producing a NPE? 
In java that will throw null pointer exception, that's why I'm asking.

emrahcom: I also tested for all possible cases in my test environment, with `null`, empty string and non-empty string

damencho: 👍 sorry my kotlin knowledge

emrahcom: It is new for me too :)
But I checked documents

damencho: I also checked but when you told me it is fine :)

damencho: I should have checked earlier ...

PR #508, opened 2023-04-24
feat: refactor, remove hardcoded youtube rtmp url

roanta2:
Refactoring on livestream part to not treat youtube as a special case. So now you need to provide the entire rtmp url ( stream url + key).

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/api/http/HttpApi.kt
+++ b/src/main/kotlin/org/jitsi/jibri/api/http/HttpApi.kt
@@ -62,7 +62,7 @@ data class StartServiceParams(
      */
     val callLoginParams: XmppCredentials? = null,
     val sinkType: RecordingSinkType,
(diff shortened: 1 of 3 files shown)


bbaldino:
Others probably have a better memory of this than I do, but there's actually existing code that I believe is trying to support exactly this.  If you look at this chunk you removed in your PR:

```kotlin
                val rtmpUrl = if (startIq.streamId.isRtmpUrl()) {
                    startIq.streamId
                } else {
                    "$YOUTUBE_URL/${startIq.streamId}"
                }
```
That's allowing for the value passed as the `streamId` to be an entire RTMP url.

saghul:
@roanta2 We'll need to modify Jitsi Meet too to use the full rtmp URL format.

roanta2:
@saghul Thanks. I'm mostly backend but it seems like a minor change in Jisti Meet, so I'll give it a try.

PR #510, opened 2023-05-16
feature: pre-warm chrome on startup

aaronkvanmeerten:


damencho (approved):

saghul:
Shouldn't the URL be configurable? This won't help other deployments, will it?

PR #518, opened 2023-10-01
docs: update the installation guide

emrahcom:
- Delete the notes that are no longer valid and replace them with the current working way.
- Format the markdown text


saghul (approved):

saghul:
🙏

PR #531, opened 2023-11-21
remove xdotool dependency

532910:
xdotool is not used more and debian package doesn't depend on it


532910:
Yep, I've signed this in the past.

saghul (approved):

PR #540, opened 2024-02-28
feat: Add prometheus metrics.

bgrozev:
- ref: Move stats config to a separate file.
- ref: Add an abstraction layer in front of statsd.
- ref: Move statsd code under metrics, make private.
- Add a prometheus interface at /metrics.
- feat: Add "healthy" and "recording" metrics.
- ref: Rename a function for clarity.


src/main/kotlin/org/jitsi/jibri/metrics/JibriMetrics.kt
```diff
@@ -0,0 +1,166 @@
...
+        NonBlockingStatsDClient(
+            "jibri",
+            StatsConfig.statsdHost,
+            StatsConfig.statsdPort
+        )
+    } else {
+        null
+    }
+
+    private fun incrementStatsDCounter(aspect: String, vararg tags: String) {
```
JonathanLennox, line 39: Does this ever actually take more than one tag String?

I see places that pass "tags: String" but that's still just a single String.

src/main/kotlin/org/jitsi/jibri/metrics/JibriMetrics.kt
```diff
@@ -0,0 +1,166 @@
...
+            "errors",
+            "Number of errors."
+        )
+        val requestsWhileBusy = JibriMetricsContainer.registerCounter(
+            "busy",
+            "Number of times a request was received while the instance was busy."
+        )
+        val xmppConnected = JibriMetricsContainer.registerCounter(
+            "xmpp_connected",
+            "Number of time an XMPP connection connected."
```
JonathanLennox, line 125: Should be "number of times"; also for a number of stats below.

JonathanLennox (approved):

PR #546, opened 2024-09-17
feat(metadata): Keeps track of participants last used display name.

damencho:


src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -141,10 +142,21 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
...
                         if (!hidden && identity) {
-                            window._jibriParticipants.push(identity);
+                            identity.lastKnownDisplayName = nick;
+                            window._jibriParticipants.set(from, identity);
                         }
                     }
                 );
+                APP.conference._room.room.addListener(
+                    "xmpp.display_name_changed",
+                    (jid, displayName) => {
```
saghul, line 152: Skip if display name is undefined or 0 length.

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -141,10 +142,21 @@ class CallPage(driver: RemoteWebDriver) : AbstractPageObject(driver) {
...
+                        const identity = window._jibriParticipants.get(jid);
+                        if (identity) {
+                            identity.lastKnownDisplayName = displayName;
```
saghul, line 155: Pl set the display name property instead, so the JaaS webhook payload doesn't change since it takes it from the metadata file.

damencho: So identity already has the display name that is coming from the jwt. Do you want to override that one and lost any values from it?

damencho: I see, to overwrite name in user:
```
{
  "user": {
    "id": "123",
    "name": "damencho",
    "avatar": "https://"
  },
  "group": "456"
}
```

saghul (approved):

PR #552, opened 2024-11-12
ref: Move jwt utils to jicoco.

bgrozev:


src/main/kotlin/org/jitsi/jibri/webhooks/v1/WebhookClient.kt
```diff
@@ -70,17 +59,7 @@ class WebhookClient(
     }
 
     // We refresh 5 minutes before the expiration
```
JonathanLennox, line 61: Comment is still technically true I guess, but not specified here.

JonathanLennox (approved):

PR #560, opened 2025-02-06
Make all ffmpeg options configurable.

bgrozev:
This makes all ffmpeg options configurable in jibri.conf. The default audio codec is switched to opus. The "-tune zerolatency" and "-g 60" parameters are enabled only for live-streaming (lowering encoding complexity for file recording)


src/main/resources/reference.conf
```diff
@@ -56,6 +63,67 @@ jibri {
...
+    // Input parameters specific to linux
+    input-linux = [
+      "-f", "x11grab"
+      "-draw_mouse", "0"
+      "-r", ${jibri.ffmpeg.framerate}
+      "-s", ${jibri.ffmpeg.resolution}
+      "-thread_queue_size", ${jibri.ffmpeg.queue-size}
+      "-i", ":0.0+0,0"
+      "-f", ${jibri.ffmpeg.audio-source}
+      "-thread_queue_size", ${jibri.ffmpeg.queue-size}
```
JonathanLennox, line 79: You have thread_queue_size listed twice

src/main/resources/reference.conf
```diff
@@ -56,6 +63,67 @@ jibri {
...
+      "-f", "avfoundation"
+      "-framerate", ${jibri.ffmpeg.framerate}
+      "-video_size" ${jibri.ffmpeg.resolution}
+      // Note the values passed here will need to be changed based on the output of
+      // ffmpeg -f avfoundation -list_devices true -i ""
+      "-i", "0:0"
+      "-vsync", "2"
+    ]
+    // Audio encoding parameters
+    audio-params = [ "-acodec", "libopus", "-b:a", "64k" ]
```
JonathanLennox, line 94: Is Opus widely-enough supported in video players?

src/main/kotlin/org/jitsi/jibri/sink/impl/FileSink.kt
```diff
@@ -29,23 +32,18 @@ import java.time.format.DateTimeFormatter
...
-    override val options: Array<String> = arrayOf(
-        "-profile:v",
-        "main",
-        "-level",
-        "3.1"
-    )
+    override val format: String = recordingFormat
 
     companion object {
+        val recordingFormat: String by config("jibri.ffmpeg.recording-format".from(Config.configSource))
```
JonathanLennox, line 46: This is both the argument to '-f' in ffmpeg and also the file extension? Do those always match?

src/main/resources/reference.conf
```diff
@@ -56,6 +63,67 @@ jibri {
...
+      "-c:v", "libx264"
+      "-preset", ${jibri.ffmpeg.video-encode-preset}
+      "-pix_fmt", "yuv420p",
+      "-r", ${jibri.ffmpeg.framerate}
+      "-crf", ${jibri.ffmpeg.h264-constant-rate-factor}
+    ]
+    // Video encoding parameters specific to file recording
+    video-params-recording = ${jibri.ffmpeg.video-params} [
+      "-profile:v", "main"
+      "-level", "3.1"
```
JonathanLennox, line 106: Why do you specify a profile and level for recording but not for streaming?

src/main/kotlin/org/jitsi/jibri/sink/impl/FileSink.kt
bgrozev: No, for example ffmpeg uses 'matroska' for the muxer but the is usually '.mkv'. I figured the extension is not important and I'd rather have the format configurable

JonathanLennox: You could have them be two separate config parameters, with one value defaulting to the other one.

src/main/resources/reference.conf
bgrozev: Yes, that's for the different sources (-i), it's been like this in jibri:
https://github.com/jitsi/jibri/blob/0271286b4c9dfbe5807a835bd1244e919be47f39/src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/Commands.kt#L28C9-L28C35

src/main/resources/reference.conf
bgrozev: No idea, I just ported the existing flags.

bgrozev: https://github.com/jitsi/jibri/blob/0271286b4c9dfbe5807a835bd1244e919be47f39/src/main/kotlin/org/jitsi/jibri/sink/impl/StreamSink.kt#L28
https://github.com/jitsi/jibri/blob/0271286b4c9dfbe5807a835bd1244e919be47f39/src/main/kotlin/org/jitsi/jibri/sink/impl/FileSink.kt#L41

JonathanLennox: Ok; while changing/rationalizing these is probably a good idea, we should keep it as a separate PR to avoid being confusing.

JonathanLennox (approved):

PR #596, opened 2026-04-10
Make finalize-script optional for file recordings

thisisnkc:
## Make finalize-script optional

The `finalize-script` in `reference.conf` had a dummy default (`/path/to/finalize`) which meant every deployment had to override it, even if they didn't need a finalize script at all.

This changes it so the property is optional i.e, if you don't set it, Jibri just skips finalization and moves on. Empty strings are also handled the same way.

No breaking changes, existing setups that already configure a finalize script will work exactly as before.


src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
```diff
@@ -116,10 +117,7 @@ class FileRecordingJibriService(
         "JibriConfig::recordingDirectory" { Config.legacyConfigSource.recordingDirectory!! }
         "jibri.recording.recordings-directory".from(Config.configSource)
     }
-    private val finalizeScriptPath: String by config {
-        "JibriConfig::finalizeRecordingScriptPath" {
-            Config.legacyConfigSource.finalizeRecordingScriptPath!!
-        }
+    private val finalizeScriptPath: String? by optionalconfig {
```
saghul, line 120: You are not handling the backwards compat here, are you?

saghul:
The idea SGTM, can't comment much on the code.

src/main/kotlin/org/jitsi/jibri/service/impl/FileRecordingJibriService.kt
thisisnkc: Good catch, thanks! I missed the legacy config source when switching to optionalconfig. Fixed now.

thisisnkc:
I missed on the backward compatibility, should we add tests for that as I can see there are none currently?

PR #597, opened 2026-04-13
feat: add audio-only recording mode to reduce CPU usage

agrille:
## Summary

Adds a configurable audio-only recording mode that skips video capture and encoding entirely. This reduces CPU and storage requirements, making Jibri viable on lower-spec machines and ideal for use cases like meeting transcription.

## Changes

- `reference.conf`: adds `enable-video` flag (default `true`) and audio-only FFmpeg input/output commands for Linux and macOS
- `FfmpegCapturer`: selects audio-only command when `enable-video = false`
- `FileSink`: uses `.m4a` extension for audio-only recordings
- Tests for both new code paths

## Usage

Set in `jibri.conf`:
jibri.ffmpeg.enable-video = false

Output files use `.m4a` extension (AAC audio in MPEG-4 container).

## Motivation

Recording a Jitsi meeting for transcription purposes does not require video.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/FfmpegCapturer.kt
+++ b/src/main/kotlin/org/jitsi/jibri/capture/ffmpeg/FfmpegCapturer.kt
@@ -57,17 +57,26 @@ class FfmpegCapturer(
         const val COMPONENT_ID = "Ffmpeg Capturer"
         private val ffmpegOutputLogger = getLoggerWithHandler("ffmpeg", FfmpegFileHandler())
 
+        val enableVideo: Boolean by config {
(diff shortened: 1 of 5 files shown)


jjfreire:
+1

mfmendez:
+1

rodrigoGrZa:
+1

Manuframil:
+1

saghul:
Folks, just to adjust expectations: posting +1 won't help get this feature landed.

Specifically, the UI has no control over this, which means there is no way to decide whether one wants an audio-only or video recording.

For that reason alone, this cannot land.

If what you are interested in is transcriptions, the righ approach is to use JMR: https://github.com/jitsi/jitsi-multitrack-recorder which gets audio directly from the JVB and is much more lightweight.

PR #598, opened 2026-04-16
fix: Don't advertise IDLE until ChromeDriver is validated at startup

aaronkvanmeerten:
## Summary

- Jibri was broadcasting `IDLE+HEALTHY` to XMPP MUCs immediately on startup before ChromeDriver was validated, allowing Jicofo to dispatch requests that would fail during ChromeDriver initialization
- Adds an `EnvironmentChecker` that registers as `UNHEALTHY` at construction, then probes ChromeDriver in the background — Jibri only becomes `HEALTHY` (eligible for dispatch) after the probe succeeds
- Retry defaults (6 attempts, 10s delay) accommodate post-reboot delays where Xvfb/Chrome may not be immediately available
- Configurable via `jibri.chrome.startup-check`, `startup-check-max-attempts`, and `startup-check-retry-delay`

## Test plan

- [x] Unit tests for `EnvironmentChecker` (initial UNHEALTHY registration, successful transition, failure persistence, retry-then-succeed)
- [x] Full `mvn verify` passes (145 tests, 0 failures, ktlint clean)
- [ ] Deploy to a test Jibri and verify:
  - Initial XMPP presence shows `IDLE + UNHEALTHY`
  - After probe succeeds (~seconds), presence updates to `IDLE + HEALTHY`
  - HTTP `/jibri/api/v1.0/health` reflects the same transitions
  - Health check script does not trigger shutdown during normal startup window
- [ ] Verify `startup-check = false` preserves old behavior (immediate `IDLE + HEALTHY`)

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/Main.kt
+++ b/src/main/kotlin/org/jitsi/jibri/Main.kt
@@ -28,6 +28,7 @@ import org.jitsi.jibri.config.Config
 import org.jitsi.jibri.config.XmppEnvironmentConfig
 import org.jitsi.jibri.config.loadConfigFromFile
 import org.jitsi.jibri.config.toXmppEnvironment
+import org.jitsi.jibri.environment.EnvironmentChecker
 import org.jitsi.jibri.status.ComponentBusyStatus
 import org.jitsi.jibri.status.ComponentHealthStatus
 import org.jitsi.jibri.status.JibriStatusManager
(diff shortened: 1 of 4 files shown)


bgrozev (changes requested):
Suggest to nest the config under a "startup-check" section:
chrome {
  startup-check {
    enabled = true
    max-attempts = 6
    retry-delay = 10000
  }
}

aaronkvanmeerten (commented):
> Suggest to nest the config under a "startup-check" section:
> chrome {
>   startup-check {
>     enabled = true
>     max-attempts = 6
>     retry-delay = 10000
>   }
> } 

Good idea, done.

bgrozev (approved):

PR #608, opened 2026-06-05
feat: macOS screen capture support in the local-run script

damencho:
Add JIBRI_MAC_VIDEO_DEVICE / JIBRI_MAC_AUDIO_DEVICE knobs that override ffmpeg's avfoundation input in the generated jibri.conf, so a screen device plus a loopback audio device (e.g. BlackHole) can be captured instead of the default camera + mic. avfoundation screen devices only deliver native (retina) resolution, so -video_size is dropped from the input and the output is scaled to the configured recording width.

Document the macOS setup (Screen Recording permission, BlackHole multi-output routing, device indices) and add troubleshooting entries for black video on macOS (missing TCC permissions) and Linux (Wayland sessions, where x11grab cannot see the screen).


saghul (approved):
Nice!

PR #609, opened 2026-06-12
feat: Add recorder pages and feature flag

hlvhe:
- CallPage interface for AppCallPage (page loading) and ExternalAPIPage (External API)
- feature flag JIBRI_USE_EXTERNAL_API


damencho:
Can you add recorder.html to this folder: https://github.com/jitsi/jibri/tree/master/src/main/resources

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -0,0 +1,86 @@
...
+class ExternalAPIPage(driver: RemoteWebDriver) : AbstractPageObject(driver), CallPage {
+    private val logger = createLogger()
+
+    init {
+        PageFactory.initElements(driver, this)
+    }
+
+    override fun visit(url: CallUrlInfo): Boolean {
+        val room = extractRoomName(url).lowercase()
+        val recorderUrl = "https://192.168.1.4:8443/recorder.html?room=$room"
```
damencho: You can extract the recorder file from the jar and save it in a tmp folder (`javaClass.getResourceAsStream("/recorder.html")`) and load it via file://, you will just need to pass baseURL encoded as param like the room, so you can use it for external_api retrieval.

src/main/resources/recorder.html
```diff
@@ -0,0 +1,99 @@
...
+                const domain = window.location.host;
+
+                console.log(`[recorder] Initializing for room=${room} domain=${domain}`);
+
+                try {
+                    const api = new JitsiMeetExternalAPI(domain, {
+                        roomName: room,
+                        parentNode: document.querySelector("#jitsiMeetContainer"),
+                        configOverwrite: {
+                            hosts: { domain, muc: `muc.${domain}` },
```
damencho: Why do you need to override hosts? That anyway is something that jitsi-meet do not allow to override. Also muc.domain is correct only for docker deployment, for those using deb packages this is conference.domain.

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -18,632 +18,42 @@
...
-    fun sendPresence(): Boolean {
-        val result = driver.executeScript(
-            """
-            try {
-                APP.conference._room.room.sendPresence();
-            } catch (e) {
-                return e.message;
+    companion object {
+        fun create(driver: RemoteWebDriver): CallPage {
+            val useExternalAPI = System.getenv("JIBRI_USE_EXTERNAL_API") == "true"
```
damencho: We should move that to the jibri.conf `reference.conf/jibri.conf` and use it from there.

src/main/resources/recorder.html
hlvhe: Updated

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
hlvhe: Updated

src/main/resources/reference.conf
```diff
@@ -134,6 +134,9 @@ jibri {
     command-mac-recording = ${jibri.ffmpeg.ffmpeg} ${jibri.ffmpeg.input-mac} ${jibri.ffmpeg.audio-params} ${jibri.ffmpeg.video-params-recording}
     command-mac-streaming = ${jibri.ffmpeg.ffmpeg} ${jibri.ffmpeg.input-mac} ${jibri.ffmpeg.audio-params} ${jibri.ffmpeg.video-params-streaming}
   }
+  selenium {
+    use-external-api = true
```
damencho: This should default to false.

src/main/resources/reference.conf
```diff
@@ -134,6 +134,9 @@ jibri {
     command-mac-recording = ${jibri.ffmpeg.ffmpeg} ${jibri.ffmpeg.input-mac} ${jibri.ffmpeg.audio-params} ${jibri.ffmpeg.video-params-recording}
     command-mac-streaming = ${jibri.ffmpeg.ffmpeg} ${jibri.ffmpeg.input-mac} ${jibri.ffmpeg.audio-params} ${jibri.ffmpeg.video-params-streaming}
   }
+  selenium {
```
damencho, line 138: Can you add a comment here saying that this is WIP.

damencho (changes requested):

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -0,0 +1,119 @@
...
+     * directly in the browser instead of relying on the deployment to serve it.
+     * Extracted lazily and reused for the lifetime of the page object.
+     */
+    private val recorderHtmlFile: File by lazy { extractRecorderHtml() }
+
+    init {
+        PageFactory.initElements(driver, this)
+    }
+
+    override fun visit(url: CallUrlInfo): Boolean {
```
damencho, line 29: The original visit returns true only when the meeting was joined, while here this is not the case.

src/main/resources/recorder.html
```diff
@@ -0,0 +1,104 @@
...
+    </head>
+    <body>
+        <div id="jitsiMeetContainer"></div>
+        <script>
+            // This page may be loaded from a local file (file://) instead of being
+            // served by the deployment, so the deployment base url is passed in as a
+            // query parameter and used to source external_api.js and to connect to.
+            const params = new URLSearchParams(window.location.search);
+            const room = params.get("room");
+            const baseUrl = params.get("baseUrl");
```
damencho: Will this take into account any tenant?
If the meeting is https://example.com/tenant/roomname?

src/main/resources/recorder.html
```diff
@@ -0,0 +1,104 @@
...
+                console.error("[recorder] Missing required 'room' parameter");
+            } else if (!baseUrl) {
+                console.error("[recorder] Missing required 'baseUrl' parameter");
+            } else {
+                const script = document.createElement("script");
+                script.src = `${baseUrl.replace(/\/$/, "")}/external_api.js`;
+                script.onload = () => initializeRecorder(room, baseUrl);
+                script.onerror = () =>
+                    console.error(`[recorder] Failed to load external_api.js from ${baseUrl}`);
+                document.head.appendChild(script);
```
damencho: onError should handle jibriPageState to indicate there is some problem with loading external API

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -0,0 +1,119 @@
...
+            logger.info("Loading recorder page for room=$room baseUrl=$baseUrl from $recorderUrl")
+            driver.get(recorderUrl)
+            true
+        } catch (t: Throwable) {
+            logger.error("Failed to load recorder page: ${t.message}", t)
+            false
+        }
+    }
+
+    private fun extractRoomName(url: CallUrlInfo): String {
```
damencho: Maybe there is no point of this wrapper?

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -0,0 +1,119 @@
...
+
+    override fun unmute() = true
+
+    override fun getNumParticipants(): Int {
+        return try {
+            val result = driver.executeScript(
+                "return window.jibriPageState ? window.jibriPageState.getRemoteParticipantCount() + 1 : 1;"
+            )
+            val count = (result as? Number)?.toInt() ?: 1
+            logger.info("getNumParticipants: result=$result, count=$count")
```
damencho: If we want to keep this, it should be debug.

damencho (changes requested):

hlvhe:
@damencho, thanks for the review! I've addressed the feedback.

src/main/resources/recorder.html
```diff
@@ -0,0 +1,104 @@
...
+				script.onerror = () => {
+					const errorMsg = `Failed to load external_api.js from ${baseUrl}`;
+					window.jibriPageState = {
+						apiError: errorMsg,
+					};
+				};
+				document.head.appendChild(script);
+			}
+
+			function initializeRecorder(room, baseUrl) {
```
damencho: This is defined but not used?

src/main/resources/recorder.html
```diff
@@ -0,0 +1,104 @@
...
+							startWithVideoMuted: true,
+							prejoinConfig: { enabled: false },
+							p2p: { enabled: false },
+						},
+					});
+
+					window.jibriRecorderApi = api;
+
+					window.jibriPageState = {
+						conferenceJoined: false,
```
damencho, line 76: Where do we set this to true?

damencho: And there is no listener for it.

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -0,0 +1,132 @@
...
+
+            WebDriverWait(driver, Duration.ofSeconds(30)).until {
+                (
+                    driver.executeScript(
+                        "return window.jibriPageState?.conferenceJoined === true;"
+                    ) as? Boolean
+                    ) ?: false
+            }
+            logger.info("Recorder page initialized successfully")
+            true
```
damencho, line 59: This will always return true? Even when not able to join?

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -0,0 +1,132 @@
...
+            logger.info("Loading recorder page for room=$room baseUrl=$baseUrl from $recorderUrl")
+            driver.get(recorderUrl)
+
+            val apiError = driver.executeScript("return window.jibriPageState?.apiError;")
+            if (apiError != null) {
+                logger.error("External API failed to initialize: $apiError")
+                return false
+            }
+
+            WebDriverWait(driver, Duration.ofSeconds(30)).until {
```
damencho, line 49: This poll for state should be checking apiError, as the check above for apiError can miss any error as it runs immediately after driver.get() before the async script.onerror

damencho (changes requested):

src/main/resources/recorder.html
hlvhe: Fixed

src/main/resources/recorder.html
hlvhe: Fixed

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
hlvhe: Added check in polling

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
hlvhe: Updated

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -0,0 +1,130 @@
...
+        val baseUrl = url.baseUrl
+        val recorderUrl = buildString {
+            append(recorderHtmlFile.toURI().toString())
+            append("?room=").append(encode(room))
+            append("&baseUrl=").append(encode(baseUrl))
+        }
+        logger.info("Loading recorder page for room=$room baseUrl=$baseUrl from $recorderUrl")
+        driver.get(recorderUrl)
+
+        val apiError = driver.executeScript("return window.jibriPageState?.apiError;")
```
damencho: Do you need this check if it will be addressed in the while below?

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -0,0 +1,130 @@
...
+        if (apiError != null) {
+            logger.error("External API failed to initialize: $apiError")
+            return false
+        }
+
+        return try {
+            WebDriverWait(driver, Duration.ofSeconds(30)).until {
+                driver.executeScript("return !window.jibriPageState?.apiError && window.jibriPageState?.conferenceJoined === true;") as? Boolean
```
damencho: If there is an error it will need 30 second to finish the while. You need to redo this peace as it always returns false after a timeout means there is no early exit like apiError ...

src/main/resources/recorder.html
damencho: > Will this take into account any tenant? If the meeting is https://example.com/tenant/roomname?

Yep, I have tested your PR with https://192.168.86.102:8443/tenant/damencho and it doesn't work.
I jibri I see `Joining MUC as damencho@muc.meet.jitsi/....` where in the web I see `Joining MUC as damencho@muc.tenant.meet.jitsi/...`.

hlvhe: Fixed

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
hlvhe: Fixed

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
hlvhe: No, removed

damencho (approved):

PR #610, opened 2026-07-01
fix: Use getRoomsInfo() for number of participants

hlvhe:
This PR adds use of `getRoomsInfo()` in `ExternalAPIPage`, counting participants from the main room.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
@@ -83,6 +83,7 @@ class ExternalAPIPage(driver: RemoteWebDriver) : AbstractPageObject(driver), Cal
             val result = driver.executeScript(
                 "return window.jibriRecorderApi ? window.jibriRecorderApi.getNumberOfParticipants() : 1;"
             )
+            logger.debug("Number of participants: $result")
             (result as? Number)?.toInt() ?: 1
         } catch (t: Throwable) {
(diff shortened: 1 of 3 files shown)


damencho (changes requested):
I think getNumberOfParticipants returns and the all the participants that are in breakout rooms. Test that.

hlvhe:
Confirmed and updated the PR description.

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -80,14 +80,25 @@ class ExternalAPIPage(driver: RemoteWebDriver) : AbstractPageObject(driver), Cal
 
     override fun getNumParticipants(): Int {
         return try {
-            val result = driver.executeScript(
-                "return window.jibriPageState ? window.jibriPageState.getRemoteParticipantCount() + 1 : 1;"
-            )
-            val count = (result as? Number)?.toInt() ?: 1
-            logger.debug("getNumParticipants: result=$result, count=$count")
+            driver.manage().timeouts().scriptTimeout(Duration.ofSeconds(20))
```
damencho: This should be set once globally not per call.

damencho (changes requested):

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
hlvhe: Moved to init.

damencho (approved):

PR #611, opened 2026-07-07
fix: fetch bitrates from conference stats

hlvhe:
Implements `ExternalAPIPage.getBitrates()` by reading bitrates from the conference iframe's `APP.conference.getStats()`. This enables the recorder to detect media.


damencho:
We should expose the stats in the iframeAPI and use it from there.

PR #614, opened 2026-07-15
fix: Restore credentials in recorder page's localStorage

hlvhe:
`recorder.html` stopped storing credentials in the page's localStorage, so Jibri joined and stayed visible.

### DIFF
--- a/src/main/resources/recorder.html
+++ b/src/main/resources/recorder.html
@@ -56,6 +56,17 @@
 				);
 
 				try {
+					if (config.appData?.localStorageContent) {
+						const decoded = decodeURIComponent(
+							config.appData.localStorageContent,
+						);
+						const localStorageContent = JSON.parse(decoded);
+						localStorage.setItem(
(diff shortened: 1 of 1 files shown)


damencho:
This means that something else is wrong. You should be able to pass the credentials. Current non external-api flow, does not store anything in the local storage and pass them via appdata param, that should also work with iframe API. You were setting use host local stoarge to true, maybe that is causing the issue.

hlvhe:
Thanks for the review. Wasn't able to pass the credentials via the config (`appData.localStorageContent`) instead of local storage. In my testing, credentials only reach the iframe when `useHostPageLocalStorage` is true and `localStorage.setItem()` is called in `recorder.html`. Proposal: hardcode `useHostPageLocalStorage: true`, so it's always set?

damencho:
I just tested the latest master and the recorder is hidden and that part is working fine.

damencho (approved):

PR #620, opened 2026-07-28
feat(external-api): Query local audio/video mute state

hlvhe:
Implements `isLocalAudioMuted()` and `isLocalVideoMuted()` in `ExternalAPIPage`, used by the *6/*7 DTMF mute-toggle handlers.


damencho (changes requested):
The old code fallbacks to muted in case of error while here it is reversed.

hlvhe:
Fixed in [8c3ed3f](https://github.com/jitsi/jibri/pull/620/commits/8c3ed3f072863735cf8dcd75c51a2f61e6b59f78)

damencho (approved):

PR #623, opened 2026-07-30
feat(external-api): Detect local participant kicked

hlvhe:
This PR adds detection so the recorder stops recording when it gets kicked from the room.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
@@ -212,7 +212,16 @@ class ExternalAPIPage(driver: RemoteWebDriver) : AbstractPageObject(driver), Cal
         return iceConnected
     }
 
-    override fun isLocalParticipantKicked(): Boolean = false
+    override fun isLocalParticipantKicked(): Boolean {
+        return try {
+            driver.executeScript(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -212,7 +212,16 @@ class ExternalAPIPage(driver: RemoteWebDriver) : AbstractPageObject(driver), Cal
         return iceConnected
     }
 
-    override fun isLocalParticipantKicked(): Boolean = false
+    override fun isLocalParticipantKicked(): Boolean {
+        return try {
+            driver.executeScript(
+                "return window.jibriPageState.localParticipantKicked === true;"
+            ) as? Boolean ?: false
```
damencho, line 219: Should be window.jibriPageState?.localParticipantKicked === true;   ?

src/main/resources/recorder.html
```diff
@@ -101,6 +101,12 @@
 						window.jibriPageState.conferenceJoined = false;
 					});
 
+					api.addListener("participantKickedOut", ({ kicked }) => {
+						if (kicked.local) {
```
damencho: kicked?.local ?

damencho (changes requested):
Maybe we should set conferenceJoined to false when local participant is kicked?

damencho (approved):

PR #628, opened 2026-08-05
feat(external-api): Count hidden participants

hlvhe:
Gets the number of hidden participants via `getRoomsInfo(true)`, counting `isJibri === true` entries.

Depends on: https://github.com/jitsi/jitsi-meet/pull/17692

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
@@ -225,7 +225,22 @@ class ExternalAPIPage(driver: RemoteWebDriver) : AbstractPageObject(driver), Cal
         return numJigasiParticipants
     }
 
-    override fun numHiddenParticipants(): Int = 0
+    override fun numHiddenParticipants(): Int {
+        val result = callRecorderApiAsync(
+            "getRoomsInfo",
+            """
+            api.getRoomsInfo(true).then(roomsData => {
+                const mainRoom = (roomsData.rooms || []).find(r => r?.isMainRoom);
+                // isJibri is isHidden() || isHiddenFromRecorder()
+                const numHidden = (mainRoom?.participants || []).filter(p => p?.isJibri === true).length;
+                cb(numHidden);
+            }).catch(() => cb(0));
+            """
+        ) as? Number
+        val numHidden = result?.toInt() ?: 0
+        logger.debug("Number of hidden participants: $numHidden")
+        return numHidden
+    }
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -225,7 +225,22 @@ class ExternalAPIPage(driver: RemoteWebDriver) : AbstractPageObject(driver), Cal
...
     }
 
-    override fun numHiddenParticipants(): Int = 0
+    override fun numHiddenParticipants(): Int {
+        val result = callRecorderApiAsync(
+            "getRoomsInfo",
+            """
+            api.getRoomsInfo(true).then(roomsData => {
+                const mainRoom = (roomsData.rooms || []).find(r => r?.isMainRoom);
+                // isJibri is isHidden() || isHiddenFromRecorder()
```
damencho: So this does not get into account isHiddenFromRecorder?

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
```diff
@@ -225,7 +225,22 @@ class ExternalAPIPage(driver: RemoteWebDriver) : AbstractPageObject(driver), Cal
...
+                const numHidden = (mainRoom?.participants || []).filter(p => p?.isJibri === true).length;
```
damencho: This is not the same as isHidden. You can have jibri joining as sip-jibri which is not hidden and there is a participant thumbnail for it.

damencho (changes requested):

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
hlvhe: It does, fixed in [9904b02](https://github.com/jitsi/jibri/pull/628/commits/9904b028983f9316c2ebf0dbf3822dc4fcba172a)

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/ExternalAPIPage.kt
hlvhe: Fixed in [9904b02](https://github.com/jitsi/jibri/pull/628/commits/9904b028983f9316c2ebf0dbf3822dc4fcba172a), now uses the new isHidden field.

damencho (approved):

PR #632, opened 2026-08-11
fix(external-api): Local video recordings on Wayland

hlvhe:
Local recordings on Wayland contained no visible video: Chrome rendered on the wrong display, invisible to ffmpeg's `x11grab`

Changes:
- `JibriSelenium.kt`: if set, uses the `DISPLAY` env var.
- `run-jibri-local.sh`: forces `--ozone-platform=x11` so Chrome can't fall back to native Wayland rendering.

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
@@ -35,6 +35,7 @@ import org.jitsi.jibri.util.getLoggerWithHandler
 import org.jitsi.jibri.util.randomAlphaNum
 import org.jitsi.metaconfig.config
 import org.jitsi.metaconfig.from
+import org.jitsi.metaconfig.optionalconfig
 import org.jitsi.utils.logging2.Logger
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
```diff
@@ -182,6 +183,8 @@ class JibriSelenium(
     private val stateMachine = SeleniumStateMachine()
     private var shuttingDown = AtomicBoolean(false)
     private val chromeOpts: List<String> by config("jibri.chrome.flags".from(Config.configSource))
+    private val configuredChromeDisplay: String? by optionalconfig("jibri.chrome.display".from(Config.configSource))
```
damencho: Why do we need a new config, when there is already the jibriSeleniumOptions.display?

hlvhe: No new config needed, refactored to use the `DISPLAY` env var.

damencho (approved):

PR #633, opened 2026-08-12
Add tracing for colibri requests

JulianGCalderon:
This PR adds tracing for the handling of colibri requests.

Depends on https://github.com/jitsi/jicofo/pull/1301.

To test this, the following configuration is required
```conf
tracing {
    enabled = true
    otlp-protocol = "grpc"
    otlp-endpoint = "http://localhost:4317"
}
```

Related to the tracing backend GSoC project: https://summerofcode.withgoogle.com/programs/2026/projects/GseggmSv


src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -360,6 +413,30 @@ class XmppApi(
...
+        val span = tracer.spanBuilder("jibri.start")
+            .setParent(remoteContextFromIq(startIq))
+            .setAttribute("room", startIq.room.toString())
+            .setAttribute("session.id", startIq.sessionId)
+            .setAttribute("recording-mode", startIq.recordingMode.toString())
+            .startSpan()
+        try {
+            span.makeCurrent().use {
+                doHandleStartService(startIq, xmppEnvironment, environmentContext, serviceStatusHandler)
+            }
```
aaronkvanmeerten, line 431: This span ends before any of the interesting work happens: `JibriManager.startService` submits `jibriService.start()` to `TaskPools.ioPool` and jibri immediately replies PENDING, so the Selenium join, ffmpeg startup and the whole session run outside the span. Worse, OTel context is thread-local, so `makeCurrent()` here doesn't reach the pool thread — any child span created later inside the service would become an orphaned root.

Minimum fix is to propagate the context across the pool boundary in `JibriManager`:

```kotlin
TaskPools.ioPool.submit(Context.current().wrap { jibriService.start() })
```

But I think the natural unit here is a session-lifetime span (e.g. `jibri.session`) created as a child of this one, stored alongside `currentActiveService`, and ended from the status handler on `Finished`/`Error` — the three state machines already log every transition, which makes them natural hook points for span events.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -360,6 +413,30 @@ class XmppApi(
         xmppEnvironment: XmppEnvironmentConfig,
         environmentContext: EnvironmentContext,
         serviceStatusHandler: JibriServiceStatusHandler
+    ) {
+        val span = tracer.spanBuilder("jibri.start")
+            .setParent(remoteContextFromIq(startIq))
+            .setAttribute("room", startIq.room.toString())
+            .setAttribute("session.id", startIq.sessionId)
+            .setAttribute("recording-mode", startIq.recordingMode.toString())
+            .startSpan()
```
aaronkvanmeerten, line 427: Two small things on span construction:

- Consider `.setSpanKind(SpanKind.SERVER)` on both handlers (and the same for jicofo's `jibri.request` client side) so trace UIs render the cross-service hop correctly instead of showing internal spans.
- In the error path, `span.recordException(e)` alongside `setStatus(ERROR, ...)` gets you the stack trace as a span event rather than just the message.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
```diff
@@ -343,10 +386,20 @@ class XmppApi(
      * response with [JibriIq.Status.OFF].
      */
     private fun handleStopJibriIq(stopJibriIq: JibriIq): IQ {
-        jibriManager.stopService()
-        // By this point the service has been fully stopped
-        return stopJibriIq.createResult {
-            status = JibriIq.Status.OFF
+        val span = tracer.spanBuilder("jibri.stop")
+            .setParent(remoteContextFromIq(stopJibriIq))
+            .startSpan()
```
aaronkvanmeerten, line 394: The stop span has no attributes even though `session.id` is available on the stop IQ — useful for correlating with the start. It also doesn't `makeCurrent()`, so a future child span inside `stopService()` wouldn't attach to it.

Since this and `handleStartService` now share identical try/catch/finally boilerplate, a small helper would fix both and keep the handlers readable:

```kotlin
private fun <T> Tracer.withSpan(name: String, parent: Context, configure: SpanBuilder.() -> Unit = {}, block: () -> T): T {
    val span = spanBuilder(name).setParent(parent).setSpanKind(SpanKind.SERVER).apply(configure).startSpan()
    return try {
        span.makeCurrent().use { block() }
    } catch (e: Throwable) {
        span.recordException(e)
        span.setStatus(StatusCode.ERROR, e.message ?: "")
        throw e
    } finally {
        span.end()
    }
}
```

pom.xml
```diff
@@ -278,6 +266,17 @@
...
+        <dependency>
+            <groupId>org.jitsi</groupId>
+            <artifactId>jicoco-tracing</artifactId>
+            <version>${jicoco.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-sdk</artifactId>
+            <version>1.48.0</version>
+        </dependency>
```
aaronkvanmeerten: The jibri code in this PR only touches `opentelemetry-api`/`opentelemetry-context`, which come transitively from `jicoco-tracing` — the direct `opentelemetry-sdk` dependency isn't needed yet (it becomes justified once jibri needs to flush/shutdown the SDK on exit, see review body). Either way, pinning `1.48.0` here risks skewing from the OTel version `jicoco-tracing` was built against — importing `io.opentelemetry:opentelemetry-bom` in `dependencyManagement` and dropping the explicit version would keep them aligned. (Also a stray blank line after this block.)

src/main/resources/reference.conf
```diff
@@ -197,3 +197,8 @@ jibri {
     ice-connection-timeout = 30 seconds
   }
 }
+
+tracing {
+  enabled = false
+  service-name = "jibri"
+}
```
aaronkvanmeerten, line 204: `TracingConfig` *requires* `otlp-endpoint` and `otlp-protocol` once `enabled = true` (they're `onlyIf(enabled)` with no defaults), so flipping `enabled` alone throws a `ConfigException` lazily at first span creation. Suggest adding defaults so enablement is a one-liner:

```hocon
tracing {
  enabled = false
  service-name = "jibri"
  otlp-protocol = "grpc"
  otlp-endpoint = "http://localhost:4317"
}
```

aaronkvanmeerten (commented):
Thanks for this — the overall approach (remote parent from the IQ extension, mirroring the jicofo side) looks right to me. A few observations beyond the inline comments:

**The WIP "context is not being read correctly" issue is most likely already fixed.** `JibriIqProvider` used to drop all child extensions when parsing, so the `TraceParent` never made it onto the parsed `JibriIq`. That was fixed in jitsi/jitsi-xmpp-extensions#147 ("Fix Jibri IQ extension parsing"), and the version this PR pins (`1.0-118-g64ecd07`) is exactly that commit — the WIP commit predates the bump. Worth an end-to-end retest and squashing the WIP note out of the history.

**Spans will be silently dropped at process exit — worst in single-use mode.** `TracingGlobal` uses a `BatchSpanProcessor` (default 5s export delay) and never flushes; jibri exits via `exitProcess()` in `cleanupAndExit` (`Main.kt`). With `jibri.single-use-mode = true` the process exits right after the session ends — exactly when the most valuable spans were just created, so the `jibri.stop` span (and any session spans) will routinely be lost. Ideally `jicoco-tracing` would expose a `shutdown()`/`forceFlush()` (the `sdk` field is typed as the `OpenTelemetry` interface, so callers can't reach the `SdkTracerProvider` without a cast today); short of that, a JVM shutdown hook in `Main` that casts to `OpenTelemetrySdk` and closes it works.

**On the jicofo side (jitsi/jicofo#1301):** in the pending-timeout stop path, the `TraceParent` is added to `startIq` instead of `stopRequest` (copy-paste). Also both the timeout-stop and retry paths run on timer threads where `Context.current()` is root, so those stops will never carry a trace context — only user-initiated stops handled inside the `jibri.request` span will. Worth flagging over there.

**Design note on the wire format (non-blocking):** the extension carries `trace_id`/`parent_id`/`trace_flags` as separate attributes and drops `tracestate` entirely (`TraceState.getDefault()` on the receive side). Carrying the full W3C `traceparent` string (plus optional `tracestate`) in the extension and reusing `W3CTraceContextPropagator` with a trivial `TextMapGetter` over the IQ would eliminate the hand-rolled parse/build on both ends and stay spec-aligned. Might be worth settling before the format ossifies across jicofo/jibri/jvb. Relatedly, `TracingGlobal` builds the SDK with no propagators configured, which any future HTTP-header-based propagation (e.g. the ktor instrumentation on jibri's HTTP API) would silently trip over — adding `W3CTraceContextPropagator` there would future-proof it.

I'm putting together a follow-up branch with prototypes for the suggestions here (session-lifetime span with state-machine events, context propagation into the async service start, flush-on-exit, OTLP config defaults, and traceparent extraction on the HTTP API) and will link it from this PR.

aaronkvanmeerten:
As promised in my review, I've pushed a branch with prototypes for the suggestions: https://github.com/jitsi/jibri/compare/tracing-jibri...tracing-jibri-prototypes

It's based on this PR's head, with one commit per suggestion so they can be cherry-picked individually:

- `chore: Align OpenTelemetry artifacts with jicoco-tracing via the BOM` — pins the OTel version once via `opentelemetry-bom` instead of a hardcoded version on the sdk dependency.
- `ref: Extract a withSpan tracing helper, enrich the XMPP IQ spans` — new `util/Tracing.kt` with a `Tracer.withSpan` helper (makeCurrent + recordException + error status + guaranteed end); both IQ handlers use it with `SpanKind.SERVER`, and the stop span gains `session.id`.
- `feat: Add a session-lifetime span` — a `jibri.session` span in `JibriManager` that lives from the start request until the service reaches a terminal state, records state transitions as span events, and is explicitly carried across the `TaskPools.ioPool` boundary with `Context.wrap()` so future child spans (selenium join, ffmpeg) can attach. The blocking service stop is also wrapped so stop traces show teardown/finalize duration.
- `fix: Flush pending trace spans on exit` — shutdown hook closing the `OpenTelemetrySdk`; without it single-use-mode exits drop the end-of-session spans. (Longer-term this probably belongs on `TracingGlobal` itself as a `shutdown()`, which would also let jibri drop the direct sdk dependency.)
- `feat: Add OTLP defaults so tracing can be enabled with one setting` — `otlp-protocol`/`otlp-endpoint` defaults in `reference.conf`.
- `feat: Trace HTTP API start/stop requests` — extracts a remote parent from the W3C `traceparent` header, mirroring the XMPP path (done manually rather than via the ktor OTel instrumentation since `TracingGlobal` currently configures no propagators).

Compile, ktlint, and the full test suite pass on the branch. Feel free to cherry-pick whatever's useful into this PR.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
JulianGCalderon: This PR only adds spans for starting/stopping Jibri. Eventually, when we add more spans, we would need to do something like this, but I think we can leave it out of this PR, to keep it simpler.

src/main/resources/reference.conf
JulianGCalderon: To keep consistency with the other configurations (jicofo, jvb), I would keep it as it is now.

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
JulianGCalderon: Done

src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt
JulianGCalderon: I think it's a bit too soon for that utility function, the added code is not large and it's selfcontained. I would keep it as is, and consider a good utility function when we add more instrumentation.

pom.xml
JulianGCalderon: Done!

bgrozev (approved):

bgrozev:
jenkins, add to whitelist

PR #634, opened 2026-08-19
fix(external-api): Use useRawKeys when setting participant properties

hlvhe:
- ExternalAPIPage passes `useRawKeys = true` internally
- Fixes the recording notification sound not playing when Jibri uses ExternalAPIPage

Depends on: https://github.com/jitsi/jitsi-meet/pull/17735

### DIFF
--- a/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
+++ b/src/main/kotlin/org/jitsi/jibri/selenium/JibriSelenium.kt
@@ -290,9 +290,8 @@ class JibriSelenium(
 
     fun sendPresence(): Boolean = callPage.sendPresence()
 
-    fun setParticipantProperties(properties: Map<String, String>): Boolean = callPage.setParticipantProperties(
-        properties
-    )
+    fun setParticipantProperties(properties: Map<String, String>): Boolean =
+        callPage.setParticipantProperties(properties)
 
     fun handleDtmfStar6() {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
```diff
@@ -45,7 +45,7 @@ interface CallPage {
     fun raiseHand(): Boolean
     fun addToPresence(key: String, value: String): Boolean
     fun sendPresence(): Boolean
-    fun setParticipantProperties(properties: Map<String, String>): Boolean {
+    fun setParticipantProperties(properties: Map<String, String>, useRawKeys: Boolean = false): Boolean {
```
damencho: We do not need this. As useRawKeys will be used only in recorder.html that uses externa-api which will just set tru there. No need for this param in the kotlin code.

damencho (changes requested):

damencho (approved):

src/main/kotlin/org/jitsi/jibri/selenium/pageobjects/CallPage.kt
hlvhe: Removed in [d1819c2](https://github.com/jitsi/jibri/pull/634/commits/d1819c216f4a8a97aa506a83940fab8346415420)

PR #637, opened 2026-09-08
feat: Refuse an invalid start request before the instance goes busy

bgrozev:
Jibri detected a bad RTMP URL only after selenium joined the call and ffmpeg failed. By then the instance had published BUSY, so the session had to be stopped, and in single-use mode the instance was recycled. A client which repeated such a request consumed one instance per attempt, which can empty a small pool.

The request is now checked in `JibriManager` before the instance goes busy, for all three modes. No browser starts, no ffmpeg starts, and there is nothing to clean up. `StartRequestValidator` is where a check for a new parameter belongs.

A parameter which is absent must always be valid. Jicofo builds the request it sends from a fixed list of fields, so a release which does not know a parameter omits it. A test covers this.

The RTMP URL must parse, use rtmp or rtmps, and have a host and a stream key. It must match `jibri.streaming.rtmp-allow-list`, a check which moved here from `StreamingJibriService`. When the URL is the default YouTube ingest URL, the stream key must match the new `jibri.streaming.youtube-stream-key-pattern`.

How the XMPP API refuses a request depends on the requester:

- With `supports_bad_request`, it answers the start IQ with `status=off`, `should_retry=false` and a `bad-request` element which holds the reason.
- Without it, it answers `pending` and then reports the failure on the asynchronous path, exactly as a failed session does. Every release already handles this by giving up on the request. Answering such a release synchronously would make it mark this healthy instance as failed and retry the same doomed request with other instances.

The HTTP API answers 400.

Note that a well-formed key which the RTMP server rejects anyway, for example a revoked one, is still found at runtime and still costs one instance.

Depends on jitsi/jitsi-xmpp-extensions#152 (same branch name, so CI builds them together). Related: jitsi/jicofo#1316


damencho (commented):
Reviewed locally. `mvn verify` passes on the PR head — 4 new test classes, all green, checkstyle and ktlint clean. The design is right: validating in `JibriManager` before `throwIfBusy`/`publishStatus(BUSY)` is the correct place, and the "an absent parameter must always be valid" regression test is exactly the right guard.

### 1. The pending/OFF ordering in the non-opt-in path is a race, not a guarantee

`XmppApi.kt` (`rejectBadRequest`):

```kotlin
// Submitted to the io pool so that the 'pending' result is sent first...
TaskPools.ioPool.submit { serviceStatusHandler(ComponentState.Error(BadRequest(e.detail))) }
startJibriIq.createResult { status = JibriIq.Status.PENDING }
```

`ioPool` is `Executors.newCachedThreadPool()`, and the `pending` result isn't sent by Smack until `handleJibriIq` *returns*. Nothing orders the two. The comment asserts an invariant the code doesn't provide, and the test can't catch it: `inPlaceExecutor` runs `submit` inline, so in the test the OFF stanza is actually sent *before* `createResult(PENDING)` is even constructed — the opposite of the intent. If OFF wins the race in production, jicofo sees a failure for a session it hasn't yet marked pending, which is close to the confusion the PR is trying to avoid.

### 2. `badRequest()` hardcodes the live-stream statsd tag

`JibriMetrics.kt`:

```kotlin
fun badRequest() { incrementStatsDCounter(ASPECT_BAD_REQUEST, TAG_SERVICE_LIVE_STREAM) ... }
```

It's called from the generic `JibriManager.validate()`, which is on the file-recording and SIP paths too. Every sibling (`start`, `stop`, `error`, `requestWhileBusy`) takes a `RecordingSinkType`. Inert today since only streaming validates, but wrong the moment a check lands in the other two — which `StartRequestValidator`'s placeholders explicitly invite.

### 3. New unconditional structural checks reject URLs that used to work, with no way to turn them off

`RtmpUrlValidator.validate`. I ran the validator against these:

| URL | Result |
|---|---|
| `rtmp://my_server/live/abcd-1234-efgh-5678` | REJECTED: "has no host" |
| `rtmp://myserver.local/mystreamkey` | REJECTED: "has no stream key" |

The first is Java's `URI` quirk — `getHost()` returns `null` for an underscore hostname (`getAuthority()` still gives `my_server`), and Docker/internal RTMP hosts routinely have one. The second is the `pathSegments.size < 2` rule; not every RTMP server uses app/key.

`rtmp-allow-list` defaults to `.*`, so these deployments never opted into any restriction, and unlike the YouTube pattern these checks have no config knob. Suggest falling back to `uri.authority` when `host` is null, and either relaxing or making the >=2-segment rule configurable.

### 4. The YouTube key check is a literal string prefix, so the common forms bypass it

`rtmpUrl.startsWith("$YOUTUBE_URL/")`. All of these were ACCEPTED with the bad key `oooo`:

- `rtmps://a.rtmp.youtube.com/live2/oooo` — this is the form YouTube's dashboard hands out today
- `rtmp://a.rtmp.youtube.com:1935/live2/oooo`
- `RTMP://a.rtmp.youtube.com/live2/oooo`
- `rtmp://b.rtmp.youtube.com/live2/oooo` (backup ingest)

The jicofo path is covered, since `XmppApi` builds the URL from `YOUTUBE_URL` itself. But any client that sends a full URL still burns an instance on a malformed key — the exact cost the PR set out to eliminate. Matching on host + first path segment would close it.

### 5. The HTTP API feeds `youTubeStreamKey` straight in as `rtmpUrl`

`HttpApi.handleStartService`, `RecordingSinkType.STREAM` branch — no `YOUTUBE_URL` prefix, unlike `XmppApi`. A caller sending what the field name says — a bare key — now gets a hard 400 (`scheme must be rtmp or rtmps, but it is null`) where it previously got a session that failed later in ffmpeg. The naming lie is pre-existing and the 400 is arguably more honest, but the PR converts it into a hard break on an untested path. Worth a deliberate decision either way.

### Minor

- `RtmpUrlValidator.validate` KDoc says "Throws `[InvalidRtmpUrlException]`"; that class doesn't exist. It throws `BadRequestException`.
- `JibriManager.startStreaming` logs "Starting a stream" *before* validating, so a rejected request logs a start line for a session that never starts. `startFileRecording` validates first. Also, `validate()` and `rejectBadRequest()` both log the same rejection.
- The PR is careful that `BadRequestException.detail` never carries a stream key (with a test for it), but `XmppApi` already logs the full RTMP URL including the key at `Using RTMP URL $rtmpUrl ...`. Pre-existing, but it undercuts the stated invariant.
- Moving the allow-list into a companion-object `by config` means it's now resolved once at `JibriManager` construction instead of per session. Config reload does a graceful restart, so likely fine — just a change in when the value is bound.

damencho (approved):

